PR #30, opened 2022-12-16
Fix analyze map body validation

valyria257:
### Proposed changes

The current validation of a map block would check that all parameters had exactly 1 argument. This would cause parse errors on valid configs with other "map-like" directives and "special parameters". For example, the `volatile` parameter can be used without any additional arguments inside the [`map` directive](http://nginx.org/en/docs/http/ngx_http_map_module.html#map). 
Different directives also have different constraints on how many arguments a parameter can have, for example, parameter's in [`match`](http://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#match) can have 1 or more arguments.
```
http {
    match check {
        require $variable ...;
    }
}
```

This improves the previous rudimentary validation to use the bit masks, so each directive can be customized so their bodies can have different validation checks.

### Checklist

### DIFF
--- a/analyze_map.go
+++ b/analyze_map.go
@@ -1,0 +1,88 @@
+/**
+ * Copyright (c) F5, Inc.
+ *
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package crossplane
+
+import "fmt"
+
+// mapParameterMasks holds bit masks that define the behavior of the body of map-like directives.
+// Some map directives have "special parameter" with different behaviors than the default.
+type mapParameterMasks struct {
+	specialParameterMasks map[string]uint
+	defaultMasks          uint
+}
+
+//nolint:gochecknoglobals
+var mapBodies = map[string]mapParameterMasks{
+	"charset_map": {
+		defaultMasks: ngxConfTake1,
+	},
+	"geo": {
+		specialParameterMasks: map[string]uint{"ranges": ngxConfNoArgs, "proxy_recursive": ngxConfNoArgs},
+		defaultMasks:          ngxConfTake1,
+	},
+	"map": {
+		specialParameterMasks: map[string]uint{"volatile": ngxConfNoArgs, "hostnames": ngxConfNoArgs},
+		defaultMasks:          ngxConfTake1,
+	},
+	"match": {
+		defaultMasks: ngxConf1More,
+	},
+}
+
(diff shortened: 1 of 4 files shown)


analyze_map.go
```diff
@@ -0,0 +1,89 @@
...
+		return &ParseError{
+			What: fmt.Sprintf(`unexpected "%s"`, term),
+			File: &fname,
+			Line: &parameter.Line,
+		}
+	}
+
+	if mask, ok := masks.specialParameterMasks[parameter.Directive]; ok {
+		// use mask to check the parameter's arguments
+		if ((mask>>len(parameter.Args)&1) != 0 && len(parameter.Args) <= 7) || // NOARGS to TAKE7
```
dhurley, line 58: This same if statement is used twice in this function. Could you move the if condition into its own function to remove the code duplication? Doing this would also reduce the cognitive complexity of the analyzeMapBody function meaning that you would be able to remove the comment `nolint:gocyclo,gocognit` at the top of the function.

analyze_map.go
```diff
@@ -0,0 +1,89 @@
...
+	defaultMasks          uint
+}
+
+//nolint:gochecknoglobals
+var mapBodies = map[string]mapParameterMasks{
+	"charset_map": {
+		defaultMasks: ngxConfTake1,
+	},
+	"geo": {
+		specialParameterMasks: map[string]uint{"ranges": ngxConfNoArgs, "proxy_recursive": ngxConfNoArgs},
```
oliveromahony, line 25: can you add test case(s) for proxy_recursive?

analyze_map.go
```diff
@@ -0,0 +1,89 @@
...
+		defaultMasks:          ngxConfTake1,
+	},
+	"map": {
+		specialParameterMasks: map[string]uint{"volatile": ngxConfNoArgs, "hostnames": ngxConfNoArgs},
```
oliveromahony, line 29: can you add test case(s) for hostnames?

analyze_map.go
valyria257: Moved the condition to the new`hasValidArguments` helper and removed the `nolint` comment.

analyze_map.go
valyria257: Added!

analyze_map.go
valyria257: Added!

oliveromahony (approved):

dhurley (approved):

PR #31, opened 2023-01-25
add auth_jwt_require directive

xynicole:
### Proposed changes

Added [auth_jwt_require](https://nginx.org/en/docs/http/ngx_http_auth_jwt_module.html#auth_jwt_require) directive

### Checklist

### DIFF
--- a/analyze.go
+++ b/analyze.go
@@ -2074,6 +2074,9 @@ var directives = map[string][]uint{
 	"auth_jwt_type": {
 		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLmtConf | ngxConfTake1,
 	},
+	"auth_jwt_require": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLmtConf | ngxConf1More,
+	},
 	"f4f": {
 		ngxHTTPLocConf | ngxConfNoArgs,
 	},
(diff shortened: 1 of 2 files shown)


nickchen (approved):

analyze.go
```diff
@@ -2074,6 +2074,9 @@ var directives = map[string][]uint{
 	"auth_jwt_type": {
 		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLmtConf | ngxConfTake1,
 	},
+	"auth_jwt_require": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLmtConf | ngxConfTake1,
```
valyria257, line 2078: Can you check and see how many parameters you can use with `auth_jwt_require`? From the docs, it looks like more than one.

valyria257 (changes requested):

analyze.go
xynicole: you're right, it takes more than 1 args. so I changed to `ngxConf1More,`

valyria257 (approved):

analyze_test.go
```diff
@@ -119,6 +119,50 @@ func TestAnalyze_auth_jwt(t *testing.T) {
 	}
 }
 
+func TestAnalyze_auth_jwt_require(t *testing.T) {
+	t.Parallel()
+	testcases := map[string]struct {
```
ornj, line 124: For completeness should there be a test to cover `auth_jwt_require $var error=403`?

ornj (approved):

analyze_test.go
xynicole: done, added a test for the error code

oliveromahony (approved):

PR #34, opened 2023-02-22
Use golangci-lint action and update linters

lucacome:
- Use `golangci-lint-action` that can annotate PRs
- Remove deprecated linters
- Update golangci-lint in `Makefile`


.golangci.yml
```diff
@@ -25,6 +25,11 @@ linters:
     - golint
     - maligned
     - scopelint
+    # deprecated
```
lucacome, line 28: I would actually do the opposite, use `disable-all` and enable only the ones that make sense for the project...

oliveromahony (approved):

Dean-Coakley (approved):

PR #43, opened 2023-04-04
fix: validate body of types directive

valyria257:
### Proposed changes

`types` is a "map-like" directive meaning its body does not contain additional NGINX directives. Previously, if the body of the `types` directive was not empty, crossplane would fail to parse it because it did not recognize its contents as NGINX directives. This fixes this behavior by adding `types` to the set of map-like directives and validating only that each entry has the correct number of arguments, 1 or more.

For an example of the `types` directive, see http://nginx.org/en/docs/http/ngx_http_core_module.html#types.

### Checklist

### DIFF
--- a/analyze_map_test.go
+++ b/analyze_map_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// nolint:funlen
+//nolint:funlen,exhaustruct
 func TestAnalyzeMapBody(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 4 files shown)


oliveromahony (approved):

analyze_map_test.go
```diff
@@ -82,6 +82,16 @@ func TestAnalyzeMapBody(t *testing.T) {
 			},
 			term: ";",
 		},
+		"valid types": {
```
ornj, line 85: Do we need a negative test to ensure this type doesn't allow "special parameters"?

ornj (approved):

analyze_map_test.go
valyria257: added negative test case

PR #57, opened 2023-06-26
fix: Include parsed comments between arguments only when ParseComment s is true

rshyamsu:
Comments between arguments should be added to the parsed result only if the ParseComments ParseOption is true. This check was previously missing and has been added as part of this fix.

A unit test has also been added to validate this.

### Proposed changes

### Checklist


parse.go
```diff
@@ -398,15 +398,17 @@ func (p *parser) parse(parsing *Config, tokens <-chan NgxToken, ctx blockCtx, co
...
-		for _, comment := range commentsInArgs {
-			comment := comment
-			parsed = append(parsed, &Directive{
-				Directive: "#",
-				Line:      stmt.Line,
-				Args:      []string{},
-				File:      fileName,
-				Comment:   &comment,
-			})
+		if p.options.ParseComments {
```
ornj, line 401: It might be more efficient to do this check above to avoid filling `commentsInArgs` when `!ParseComments`. Did you try doing that?

rshyamsu: That makes sense. I just added the check much earlier instead of here.

ornj (approved):

oliveromahony (approved):

valyria257 (approved):

PR #62, opened 2023-08-11
fix: update nginx config file parsing logic to handle variables with …

sylwang:
…braces

### Proposed changes

Address 43818 so that the user can include variables with braces in the config file.

- update `lex.go` to correctly parse variables like `${foo}`
- add an `nginx.config` file in testdata for a test case of variables with braces
- add a test case in `parse_test.go` for checking the parser can correctly parse variables with braces
- fix the messy test case in `lex_test.go` since the values that are separated by spaces need to be tokenized 
- fix the existing linter issues, which includes updating the deprecated functions, improving readability of variable names, etc.

### Checklist

### DIFF
--- a/analyze_map_test.go
+++ b/analyze_map_test.go
@@ -178,13 +178,13 @@ func TestAnalyzeMapBody(t *testing.T) {
 		},
 	}
 
-	for name, test := range testcases {
-		testcase := test
+	for name, tc := range testcases {
+		tc := tc
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			err := analyzeMapBody("nginx.conf", testcase.parameter, testcase.term, testcase.mapDirective)
-			if testcase.wantErr == nil {
+			err := analyzeMapBody("nginx.conf", tc.parameter, tc.term, tc.mapDirective)
+			if tc.wantErr == nil {
 				require.NoError(t, err)
 				return
 			}
@@ -193,8 +193,8 @@ func TestAnalyzeMapBody(t *testing.T) {
 
 			var perr *ParseError
 			require.ErrorAs(t, err, &perr)
-			require.Equal(t, testcase.wantErr.What, perr.What)
-			require.Equal(t, testcase.wantErr.BlockCtx, perr.BlockCtx)
+			require.Equal(t, tc.wantErr.What, perr.What)
+			require.Equal(t, tc.wantErr.BlockCtx, perr.BlockCtx)
 		})
 	}
 }
--- a/analyze_map_test.go
+++ b/analyze_map_test.go
@@ -178,13 +178,13 @@ func TestAnalyzeMapBody(t *testing.T) {
 		},
 	}
 
-	for name, tc := range testcases {
-		tc := tc
+	for name, test := range testcases {
+		testcase := test
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			err := analyzeMapBody("nginx.conf", tc.parameter, tc.term, tc.mapDirective)
-			if tc.wantErr == nil {
+			err := analyzeMapBody("nginx.conf", testcase.parameter, testcase.term, testcase.mapDirective)
+			if testcase.wantErr == nil {
 				require.NoError(t, err)
 				return
 			}
@@ -193,8 +193,8 @@ func TestAnalyzeMapBody(t *testing.T) {
 
 			var perr *ParseError
 			require.ErrorAs(t, err, &perr)
-			require.Equal(t, tc.wantErr.What, perr.What)
-			require.Equal(t, tc.wantErr.BlockCtx, perr.BlockCtx)
+			require.Equal(t, testcase.wantErr.What, perr.What)
+			require.Equal(t, testcase.wantErr.BlockCtx, perr.BlockCtx)
 		})
 	}
 }
--- a/build.go
+++ b/build.go
@@ -25,7 +25,7 @@ type BuildOptions struct {
 
 const MaxIndent = 100
 
-// nolint:gochecknoglobals
+//nolint:gochecknoglobals
 var (
 	marginSpaces = strings.Repeat(" ", MaxIndent)
 	marginTabs   = strings.Repeat("\t", MaxIndent)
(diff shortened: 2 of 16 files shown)


build_test.go
```diff
@@ -544,17 +544,17 @@ func TestBuildInto(t *testing.T) {
 		t.Run(fixture.name, func(t *testing.T) {
 			t.Parallel()
 			fixture := fixture
-			sc := new(StringsCreator)
+			creator := new(StringsCreator)
```
nickchen, line 547: un-needed changes, unless it was flagged?

parse_bench_test.go
```diff
@@ -38,15 +37,15 @@ func getLargeConfig(b *testing.B) (string, func()) {
...
+	if err != nil {
 		b.Skip("cannot open input file")
 	}
-	defer f.Close()
+	defer file.Close()
 
 	// Open output file
-	tmpdir, e := ioutil.TempDir("", "alog")
-	if e != nil {
+	tmpdir, err := os.MkdirTemp("", "alog")
```
nickchen, line 47: inside tests, you can use `t.TmpDir()` without error, or `b.TmpDir()` in this case

types.go
```diff
@@ -76,47 +76,49 @@ func equals(a, b []string) bool {
...
-	if d == a {
+//
+//nolint:cyclop
+func (d *Directive) Equal(directive *Directive) bool {
+	if d == directive {
 		// same ptr, or both nil
 		return true
 	}
-	if d == nil || a == nil {
+	if d == nil || directive == nil {
```
nickchen, line 97: `directive` too close to struct name, since we are comparing one (`d`) against another, suggest `other`.

build_test.go
valyria257: This, and most of the PR's changes, seem to be fixing complaints from the `varnamelen` linter.  `varnamelen` seems too strict to me; what are both your thoughts on disabling it?

sylwang: Yeah, the `varnamelen` linter complained about the variable names, if you guys are fine with it, I can disable it.

analyze_map_test.go
```diff
@@ -178,13 +178,13 @@ func TestAnalyzeMapBody(t *testing.T) {
 		},
 	}
 
-	for name, tc := range testcases {
-		tc := tc
+	for name, test := range testcases {
+		testcase := test
```
ornj, line 182: Suggest you use the same variable name to make it impossible to accidentally use the variable that gets reassigned in the loop. I think the short variable name `tc` was a well-enough understood pattern that you should just leave it.  Assume I'd leave the same comment else where the variable name was changed.

It seems there wasn't a reason to modify this file.

types_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestDirective_String(t *testing.T) {
-	for _, tf := range []struct {
+	for _, directives := range []struct {
```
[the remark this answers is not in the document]
sylwang: Oops, will fix this typo!

build.go
```diff
@@ -66,16 +66,16 @@ func BuildFiles(payload Payload, dir string, options *BuildOptions) error {
 			return err
 		}
 
-		f, err := os.Create(path)
+		file, err := os.Create(path)
```
ornj, line 69: I'm not sure why you are renaming existing variables. These changes do not seem to be relevant to the main change you want to make. I suggest you avoid doing this so that reviewers can focus on the change in functionality you wish to introduce and give you a higher quality review.

I too used to be bugged by the short, meaningless variables names that seem to be prevalent in Go, but [when they are short lived](https://research.swtch.com/names) it's generally OK.

parse_bench_test.go
sylwang: That's good to know. Thanks for this suggestion!

build.go
```diff
@@ -104,60 +104,60 @@ func Build(w io.Writer, config Config, options *BuildOptions) error {
...
 
-	_, err := w.Write([]byte(bodyStr))
+	_, err := writer.Write([]byte(bodyStr))
 	return err
 }
 
-//nolint:gocognit
-func buildBlock(sb io.StringWriter, parent *Directive, block Directives, depth int, lastLine int, options *BuildOptions) {
-	for i, stmt := range block {
+//nolint:cyclop
```
ornj, line 111: I think you are adding this because you modified the variable names and now are getting penalized by the linter. I don't know if reverting the variable names avoids that.

There's probably a bit of code in this project that runs afoul of the linter. I think it would be better to address those cases separately rather than disabling linting rules without a good reason. If we're going to disable them, why have them?

build_test.go
```diff
@@ -438,17 +438,17 @@ func equalPayloads(p1, p2 Payload) bool {
...
-			(e1[i].Line != nil && *e1[i].Line != *e2[i].Line) {
-			println(e1[i].Error.Error())
-			println(e2[i].Error.Error())
+	for index := 0; index < len(error1); index++ {
+		if (error1[index].File != error2[index].File) ||
+			(error1[index].Error != nil && error2[index].Error != nil && error1[index].Error.Error() != error2[index].Error.Error()) ||
+			(error1[index].Line == nil) != (error2[index].Line == nil) ||
+			(error1[index].Line != nil && *error1[index].Line != *error2[index].Line) {
+			log.Println(error1[index].Error.Error())
+			log.Println(error2[index].Error.Error())
```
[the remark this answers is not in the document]
ornj: Pet peeve. Don't do that either. Use `t.Log` or `t.Logf` for this purpose.

ornj (changes requested):
It's a bit hard for me to review with so many unrelated changes. Willing to come back and review again if necessary if those get cleaned up.

build.go
sylwang: This is for fixing one of the left-over linter issues related to function complexity and was not introduced because of the new logic changes in this PR. 

Reference for cyclop: https://github.com/bkielbasa/cyclop

But yeah I agree with you, changes for fixing the linter would be better to live in its own separate PR. Thanks for your feedback Stephen!

build.go
sylwang: This is a very helpful feedback! I'll remove all the changes related to linter issues in this PR as long as they pass the linter check in the pipeline. Thanks!

build_test.go
nickchen: i'll be ok with that, reading https://github.com/blizzy78/varnamelen, seems like it's pretty strict. We can leave it for another day.

sylwang:
Closing this pull request for now, and will reopen a PR for changes to the parsing logic after [this PR for linter related changes](https://github.com/nginxinc/nginx-go-crossplane/pull/63) gets approved.

PR #64, opened 2023-08-12
Allow parsing of geoip2 module

puneetsarna:
### Proposed changes

https://docs.nginx.com/nginx/admin-guide/security-controls/controlling-access-by-geoip/ we officially recommend using GeoIP2 for
allow/restricting based on location and should
handle it in crossplane as well.

### Checklist


ornj (approved):
Does it need a test to cover `geoip2` being considered a map-like directive?

puneetsarna:
> Does it need a test to cover `geoip2` being considered a map-like directive?

@ornj done. thanks for calling it out 👍

nginx-nickc:
LGTM

nginx-nickc (approved):

PR #74, opened 2023-10-09
fix: allow 'include' directive in 'if' and 'limit_except' blocks

### LINKED ISSUE #72: "include" directive is not allowed within an "if" block
The following block of code is a valid nginx configuration: 

```   
 19    server
 20    {
 21       listen 12345;
 22       location /
 23       {
 24          if ($request_method = 'OPTIONS')
 25          {
 26             include conf.d/cors_header;
 27
 28             return 204;
 29          }
 30
 31          root /usr/share/nginx/html/;
 32          try_files $uri $uri/ /en/index.html =404;
 33
 34       }
 35    }
```

Despite that, the tool detects the "include" as a non-allowed directive under the "if" context:

```
  "errors": [
    {
      "file": "test.conf",
      "line": 26,
      "error": "\"include\" directive is not allowed here in test.conf:26"
    }
  ],
```

### To reproduce

- Take the mentioned configuration, load it in an nginx environment. Verify that it succeeds and there are no deferred errors.
- Use the same config as an input for `nginx-go-crossplane` latest version, and see the above error.

### Expected behavior

The mentioned config shouldn't be throwing an error.

### Additional context

This is likely caused by the directives map config here:
https://github.com/nginxinc/nginx-go-crossplane/blob/73c76f92a13211c7b1488dfa746d9f3353b82ff8/analyze.go#L791

I think we should modify the mask to include the "http > location > if"
```
	"include": {
		ngxAnyConf | ngxConfTake1 | ngxHTTPLifConf,
	},
```

dareste:
### Proposed changes

Resolves #72.

The mask `ngxAnyConf` is defined as "any context except `server > if`, `location > if` and `location > limit_except`".

Given that this mask is only used for defining the valid contexts for the `include` directive, and given that the contexts listed above are valid for `include`, it makes sense to modify the `ngxAnyConf` definition to include such contexts.

### Checklist

### DIFF
--- a/analyze.go
+++ b/analyze.go
@@ -55,10 +55,10 @@ const (
 )
 
 // helpful directive location alias describing "any" context
-// doesn't include ngxHTTPSifConf, ngxHTTPLifConf, or ngxHTTPLmtConf.
 const ngxAnyConf = ngxMainConf | ngxEventConf | ngxMailMainConf | ngxMailSrvConf |
 	ngxStreamMainConf | ngxStreamSrvConf | ngxStreamUpsConf |
-	ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPUpsConf
+	ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPUpsConf |
+	ngxHTTPSifConf | ngxHTTPLifConf | ngxHTTPLmtConf
 
 // map for getting bitmasks from certain context tuples
 //


ornj:
I don't know why the original definition of this mask omitted `if` contexts and `limit_except`. This went all the way back to the [original Python](https://github.com/nginxinc/crossplane/blame/master/crossplane/analyzer.py#L49) without a useful commit message or hint as to why it was implemented this way.

ornj (approved):

puneetsarna (approved):

PR #75, opened 2023-10-09
Allow otel module directives

puneetsarna:
https://nginx.org/en/docs/ngx_otel_module.html

### Proposed changes

### Checklist

### DIFF
--- a/analyze.go
+++ b/analyze.go
@@ -1010,6 +1010,24 @@ var directives = map[string][]uint{
 		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake1234,
 		ngxStreamMainConf | ngxStreamSrvConf | ngxConfTake1234,
 	},
+	"otel_exporter": {
+		ngxHTTPMainConf | ngxConfBlock | ngxConfNoArgs,
+	},
+	"otel_service_name": {
+		ngxHTTPMainConf | ngxConfTake1,
+	},
+	"otel_span_attr": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake2,
+	},
+	"otel_span_name": {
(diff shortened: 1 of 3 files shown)


ornj:
There are already directives from dynamic modules in crossplane (notably NJS), but the point has been made in the past that perhaps these should be "add ons" rather than included in the core map in `analysis.go`. I think this has some merit as you could "customize" crossplane to be aware of whatever non-standard modules you are using.

This could allow using crossplane with unknown 3rd party modules or better support those that allow configurations that can look very different from the usual NGINX configs (Lua). 

Since there are already directives from "first party" nginx modules in crossplane, I think move forward with reviewing and merging this PR.

analyze.go
```diff
@@ -1010,6 +1010,24 @@ var directives = map[string][]uint{
...
+		ngxHTTPMainConf | ngxConfTake1,
+	},
+	"otel_span_attr": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake2,
+	},
+	"otel_span_name": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake1,
+	},
+	"otel_trace": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfFlag,
```
ornj, line 1026: https://github.com/nginxinc/nginx-otel/blob/0da0f1537ec73f5683f6ca0d08b1e3a3817577da/src/http_module.cpp#L77

```suggestion
		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake1,
```

puneetsarna: done.

ornj (approved):

valyria257 (approved):

PR #85, opened 2024-04-03
Implements NGINX App Protect v5 modules through custom parse options

ornj:
### Proposed changes

Added an API for allowing the caller of `Parse` to specify bitmasks for directives outsides the "core" set of NGINX directives. This would allow the caller to parse an NGINX configuration that contained directives from a dynamic module even if it is a unknown custom one as long as the configuration still matches the expected grammar.

As part of this I moved the NAP WAF v4 directives out of the main `directives` map before adding directives for WAF v5. The two versions of this module have a different set of directives that I did not want to layer on top of each other. I want to specifically call out that this means the v4 directives are **not** included in the main directives dictionary which means a configuration using v4 NAP directives that was parsed without error before updating Crossplane would no longer pass until the `ParseOptions` are updated to include the optional match function. cc: @yluf5 @nickchen 

The NGINXaaS team is currently working on adding support for additional versions of NGINX and dynamic modules (including Lua) for which this seems like a viable way to add support for directives outside the core directive set that do not require something heavier such as a customer lexer or parser.

### Checklist

- [ ] ~I have updated any relevant documentation ([`README.md`](https://github.com/nginxinc/nginx-go-crossplane/blob/main/README.md))~

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -53,6 +53,16 @@ type parser struct {
 	includeInDegree map[string]int
 }
 
+// MatchFunc is the signature of the match function used to identify NGINX directives that
+// can be encountered when parsing an NGINX configuration that references dynamic or
+// non-core modules. The argument is the name of a directive found when parsing an NGINX
+// configuration.
+//
+// The return value is a list of bitmasks that indicate the valid contexts in which the
+// directive may appear as well as the number of arguments the directive accepts. The
+// return value must contain at least one non-zero bitmask if matched is true.
+type MatchFunc func(directive string) (masks []uint, matched bool)
+
 // ParseOptions determine the behavior of an NGINX config parse.
 type ParseOptions struct {
 	// An array of directives to skip over and not include in the payload.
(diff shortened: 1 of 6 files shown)


parse.go
```diff
@@ -93,6 +103,12 @@ type ParseOptions struct {
 
 	// If true, checks that directives have a valid number of arguments.
 	SkipDirectiveArgsCheck bool
+
+	// MatchFuncs are called in order when an unknown or non-core NGINX directive is
+	// encountered by the parser to determine the valid contexts and argument count of the
+	// directive. Set this option to enable parsing of directives belonging to non-core or
+	// dynamic NGINX modules that follow the usual grammar rules of an NGINX configuration.
+	MatchFuncs []MatchFunc
```
nginx-nickc, line 111: might be better in term of API to only require one instead of a list? caller can do the "list" in their own implementation. Or something more optimal like a map

nginx-nickc (approved):

parse.go
ornj: Map felt worse for using the same value for multiple directives. The reason for the list is to allow the caller to specify multiple functions they may not have an ability to modify without additional boilerplate.

nginx-nickc: discussed offline, possible use case for multiple is to make it easier for existing workflow of providing callback functions for different NAP.

ornj: I expect as we want to add in directives from other dynamic modules that do not require special syntax we would add additional "match functions" instead of adding them to the big dictionary of directives. You probably wouldn't enable both versions of NAP at th the same time to parse a single config (although I guess you could?).

nginx-nickc (approved):

xynicole (approved):

PR #86, opened 2024-04-17
Add all 16 *_by_lua_block and 9 *_by_lua directives to crossplane

xynicole:
### Proposed changes

Add all 16 *_by_lua_block and 9 *_by_lua directives to crossplane along with tests.
Add Lua lexer to allow Crossplane to read Lua blocks
Add Lua built
Fix incorrect Lua testdata

How do we parse Lua:
In lexer:
1. Grab everything in the Lua block as a single token and drop the {}. Example:
`content_by_lua_block {some code}`  will be  `content_by_lua_block  "some code"; `  "some code" is an argument of directive 
2. If a user puts the same argument name as the Lua block directives' name, we will treat it as a normal argument parse. Check if it starts with { which means token depth == 0.
3. If a user puts the same upstream name as the Lua block directives' name, we will treat it as a normal argument parse. Check if the token is a directive
4. Set_by_lua_block has a return value, so we need to additionally check for this directive. Get the return value token first and then obtain the rest of the block token. 
5. If a user puts quotes around the Lua directives, we need to skip these quotes.
6. If there are braces inside the Lua block, we should retain them and not remove them as we did in # 1.

In parser (analyzer):
We want to use normal directives way to parse lua block, so 
remove `ngxConfBlock` and change `ngxConfNoArgs` to `ngxConfTake1`, for `set_by_lua_block` change to `ngxConfTake2`. Just similar to `server_name`

In builder:
Add `ExternalBuild` to handle specific directives 
Add `{ }` back to the string

### Checklist


lex.go
```diff
@@ -226,3 +318,146 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+					for isSpace(next) {
+						if !ll.s.Scan() {
+							return
+						}
+						next = ll.s.Text()
+					}
+				}
+				arg += next
+			}
+			// skip leading whitespace after the return value
```
ornj, line 383: Why is this nested in the block to handle the `set_by_lua_block` token? In the Python implementation whitespace is consumed looking for the Lua block to start with `{` regardless of the matched token.

https://github.com/nginxinc/crossplane/blob/master/crossplane/ext/lua.py#L77

lex.go
```diff
@@ -226,3 +318,146 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+			for {
+				if !ll.s.Scan() {
+					return
+				}
+				next := ll.s.Text()
+
+				if !isSpace(next) {
+					if next != "{" {
+						lineno := ll.s.Line()
+						tokenCh <- NgxToken{Error: &ParseError{File: &lexerFile, What: `unexpected "}"`, Line: &lineno}}
```
ornj, line 393:
```suggestion
						tokenCh <- NgxToken{Error: &ParseError{File: &lexerFile, What: `expected "{" to start lua block`, Line: &lineno}}
```

ornj (commented):
Just commenting for now as I want to take a little more time to look at the lexing of Lua as I'm seeing differences between this implementation and Python other than what I expected.

analyze.go
```diff
@@ -2545,54 +2545,129 @@ var LuaDirectives = map[string][]uint{
 	"lua_package_cpath": {
 		ngxHTTPMainConf | ngxConfTake1,
 	},
+	"init_by_lua": {
```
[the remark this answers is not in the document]
ornj: I missed it, you did!

lex.go
xynicole: done

lex.go
xynicole: Done

lex.go
```diff
@@ -226,3 +318,146 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+	tokenDepth := 0
+
+	go func() {
+		defer close(tokenCh)
+		var tok strings.Builder
+		var inQuotes bool
+
+		// special handling for'set_by_lua_block' directive
+		// #nosec G101
+		if matchedToken == "set_by_lua_block" { //nolint
```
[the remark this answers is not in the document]
xynicole: Done

lex.go
```diff
@@ -43,14 +43,58 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
 func SetTokenChanCap(size int) {
 	tokChanCap = size
 }
-func Lex(reader io.Reader) chan NgxToken {
+
+type LexScanner interface {
```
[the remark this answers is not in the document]
xynicole: Done

lex.go
```diff
@@ -43,14 +43,63 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
 func SetTokenChanCap(size int) {
 	tokChanCap = size
 }
-func Lex(reader io.Reader) chan NgxToken {
+
+// LexScanner provides an interface for scanning tokens.
```
ornj, line 47: Nit: `LexScanner` _is_ and interface, it doesn't provide one. Maybe something like:

```suggestion
// LexScanner is the interface used by Lex to tokenize statements or blocks that may not conform to 
// the standard NGINX grammar. 
```

You could also provide additional detail to explain further how this type should be used. [See `bufio.Scanner` as an example](https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/bufio/scan.go;l=14).

Looking at this again, I don't think that `LexScanner` needs to be an interface at all as there is only a single implementation of the interface (`*extScanner`) and that could probably be used when necessary in tests.

lex.go
```diff
@@ -43,14 +43,63 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
...
+type LexScanner interface {
+	// Scan is used to advance the scanner to the next token.
```
ornj, line 49: Nit: There's an opportunity to provide some direction for the reader to help them understand how to use this type. [Stealing some bits from `bufio.Scanner.Scan`](https://pkg.go.dev/bufio#Scanner.Scan) assuming this was a concrete type and not an interface.

```suggestion
	// Scan advances the [LexScanner] to the next token, which will then be 
	// available  through the [LexScanner.Text] method. It returns false when
	// there are no more tokens, either by reaching the end of the input or an 
	// error... (there's more that could be said about it, but you get the idea)
```

lex.go
```diff
@@ -43,14 +43,63 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
...
+	Line() int
+}
+
+type ExtLexer interface {
+	Register(scanner LexScanner) []string
+	Lex(matchedToken string) <-chan NgxToken
+}
+
+type LexOptions struct {
+	ExternalLexers []ExtLexer
```
ornj, line 65: Nit: A docstring here would help users understand what is being configured.

lex.go
```diff
@@ -63,15 +107,35 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
 	depth := 0
 	var la, quote string
 
+	// check if the lua token is directive, if token line is already processd, then this is not the directive
+	nextTokenIsDirective := make(map[int]tokenInfo)
```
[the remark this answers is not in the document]
xynicole: Changed it to `bool ` now

build.go
```diff
@@ -131,37 +132,54 @@ func buildBlock(sb io.StringWriter, parent *Directive, block Directives, depth i
...
 			_, _ = sb.WriteString(directive)
 
-			// special handling for if statements
-			if directive == "if" {
-				_, _ = sb.WriteString(" (")
-				for i, arg := range stmt.Args {
-					if i > 0 {
+			switch {
+			// special handling for'set_by_lua_block' directive
+			case directive == "set_by_lua_block":
```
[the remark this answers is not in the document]
xynicole: Sounds good! I implemented `ExternalBuild` to handle lua

lex.go
```diff
@@ -63,15 +107,35 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+		if externalLexers == nil {
+			externalLexers = make(map[string]ExtLexer)
+		}
+
+		if externalScanner == nil {
+			externalScanner = &extScanner{scanner: scanner, tokenLine: tokenLine}
+		}
+
+		for _, d := range ext.Register(externalScanner) {
+			externalLexers[d] = ext
```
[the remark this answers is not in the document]
xynicole: I added code to emit an error token and exit

lex.go
xynicole: done

lex.go
xynicole: I removed `LexScanner` and use `*extScanner` instead

build.go
```diff
@@ -18,9 +18,15 @@ import (
 )
 
 type BuildOptions struct {
-	Indent int
-	Tabs   bool
-	Header bool
+	Indent        int
+	Tabs          bool
+	Header        bool
+	ExternalBuild []ExtBuild // handle specific directives
```
ornj, line 24: Nit, pluralize the field name as it's a slice?

build.go
```diff
@@ -37,6 +43,44 @@ const header = `# This config was built from JSON using NGINX crossplane.
 
 `
 
+type LuaBuild struct{}
```
ornj, line 46: It might be nice to move the lua-specific types into a `lua.go` file to keep them organize together.

As per my other comments, please document these types/methods and maybe consider a single `Lua` struct that provide a builder and a lexer?

lex.go
```diff
@@ -43,14 +44,51 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
...
 	return tc
 }
 
-//nolint:gocyclo,funlen,gocognit
-func tokenize(reader io.Reader, tokenCh chan NgxToken) {
+func Lex(reader io.Reader) chan NgxToken {
+	return LexWithOptions(reader, LexOptions{})
+}
+
+type extScanner struct {
```
ornj, line 71: I think you need to make this an exported type with a better name as it's used as an argument in an exported interface. Maybe we also need to think about how else one might use this type and how it is constructed.

lex.go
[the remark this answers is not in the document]
ornj: I thought we talked offline about allowing it. The use case for that I think would be to provide an alternate implementation that "overrides" another extension in case you wanted it to behave differently for a subset of tokens?

I don't have a specific use case in mind though, so maybe this version is OK if other reviews prefer this strategy.

lex.go
```diff
@@ -100,6 +162,33 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
 			la = "\\" + la
 		}
 
+		if token.Len() > 0 {
+			tokenStr := token.String()
+			if ext, ok := externalLexers[tokenStr]; ok {
+				if nextTokenIsDirective {
```
ornj, line 168: Nit (micro-optimization), can this be checked first as it's the cheaper check vs looking for a key in the `externalLexers` map.

lex.go
```diff
@@ -100,6 +162,33 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+					// saving lex state before emitting tokenStr to know if we encountered start quote
+					lastLexState := lexState
+					if lexState == inQuote {
+						emit(tokenStartLine, true, nil)
+					} else {
+						emit(tokenStartLine, false, nil)
+					}
```
ornj, line 175: 🏌️ 

```suggestion
					emit(tokenStartLine, lexState == inQuote, nil)
```

lex.go
```diff
@@ -100,6 +162,33 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+
+					externalScanner.tokenLine = tokenLine
+					extTokenCh := ext.Lex(tokenStr)
+					for tok := range extTokenCh {
```
ornj, line 179: Maybe we document on the `Lex` interface method that the implementation must close this channel when it's finished and will not emit any more tokens. Failure to do so and we are stuck here "forever".

lex.go
```diff
@@ -226,3 +319,147 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+			next := ll.s.Text()
+			if err := ll.s.Err(); err != nil {
+				lineno := ll.s.Line()
+				tokenCh <- NgxToken{Error: &ParseError{File: &lexerFile, What: err.Error(), Line: &lineno}}
+			}
+
+			switch {
+			case next == "{" && !inQuotes:
+				tokenDepth++
+				if tokenDepth > 1 { // not the first open brace
```
ornj, line 419: I'm a little confused why this check was need, but I think it's because `tok.WriteString` is repeated in every case vs putting the check for `tokenDepth == 0` after the `switch` statement and then calling `tok.WriteString`.

lex.go
```diff
@@ -226,3 +319,147 @@ func tokenize(reader io.Reader, tokenCh chan NgxToken) {
...
+				inQuotes = !inQuotes
+				tok.WriteString(next)
+
+			default:
+				// Expected first token encountered to be a "{" to open a Lua block. Handle any other non-whitespace
+				// character to mean we are not about to tokenize Lua
+				// ignoring everything until first open brace where tokenDepth > 0
+				if isSpace(next) && tokenDepth == 0 {
+					continue
+				}
```
ornj, line 452: Can you explain what is meant by "we are not about to tokenize Lua"? I'm not sure what this is for exactly.

lex.go
xynicole: done

lex.go
xynicole: done

build.go
xynicole: done

build.go
```diff
@@ -18,9 +18,15 @@ import (
...
-	Tabs   bool
-	Header bool
+	Indent        int
+	Tabs          bool
+	Header        bool
+	ExternalBuild []ExtBuild // handle specific directives
+}
+
+type ExtBuild interface {
+	ExtDirectives() []string
```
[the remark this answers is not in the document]
xynicole: done

build.go
```diff
@@ -18,9 +18,15 @@ import (
...
+	BuildLuaBlock(sb io.StringWriter, directive string, stmt *Directive)
```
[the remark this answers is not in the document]
xynicole: done

build.go
```diff
@@ -18,9 +18,15 @@ import (
...
-	Indent int
-	Tabs   bool
-	Header bool
+	Indent        int
+	Tabs          bool
+	Header        bool
+	ExternalBuild []ExtBuild // handle specific directives
+}
+
+type ExtBuild interface {
```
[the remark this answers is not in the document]
xynicole: done

build.go
```diff
@@ -37,6 +43,44 @@ const header = `# This config was built from JSON using NGINX crossplane.
...
+		"ssl_certificate_by_lua_block",
+		"ssl_session_fetch_by_lua_block",
+		"ssl_session_store_by_lua_block",
+	}
+}
+
+func (lb *LuaBuild) BuildLuaBlock(sb io.StringWriter, directive string, stmt *Directive) {
+	// special handling for'set_by_lua_block' directive
+	if directive == "set_by_lua_block" {
+		_, _ = sb.WriteString(" ")
```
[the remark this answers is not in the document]
xynicole: done

lex.go
ornj: Oh, ok! Maybe if the comment called out the case if it's that specific it would be clearer?

build.go
xynicole: done

lex.go
xynicole: done, change it to SubScanner

lex.go
xynicole: done

build.go
```diff
@@ -18,9 +19,22 @@ import (
...
+// can handle external and custom NGINXdirectives during the build process of NGINX configurations
+// from JSON payloads.
+type ExternalBuilder interface {
+	// RegisterExternalBuilder allows the build system to identify which NGINX directives are supported
+	// by the external builder and routes the build process of those directives to this builder.
+	RegisterExternalBuilder() []string
+	// Build is responsible for constructing the configuration block for a specific directive.
+	// It is called during the configuration building process whenever a registered directive is encountered.
+	Build(sb io.StringWriter, stmt *Directive) error
 }
```
ornj, line 38:
```suggestion
// Builder is the interface implemented by types that can render a Directive
// as it appears in NGINX configuration files.
//
// RegisterBuilder returns the names of the directives for which the builder can
// build NGINX configuration.
//
// Build writes the strings that represent the Directive and it's Block to the
// io.StringWriter returning any error encountered that caused the write to stop
// early. Build must not modify the Directive.
type Builder interface {
	RegisterBuilder() []string
	Build(sb io.StringWriter, stmt *Directive) error
}
```

- Shortened the name of the type and it's methods
- Documented usage a bit in the `interface`'s docstring (I see it this way in the Go stdlib, I think it's a bit more discoverable outside of the package this way)
- Added a warning about modifying the Directive
- Removed the mention of "JSON" as I think that's not relevant to this type

Questions about using this type:

- What happens when a directive name is not unique?
- What should happen if a different module/builder specifies directives that are in this `Directive.Block`? I don't have a specific example of this, maybe it's not a concern and this is good enough

build.go
```diff
@@ -131,37 +146,55 @@ func buildBlock(sb io.StringWriter, parent *Directive, block Directives, depth i
...
+				for _, ext := range options.ExternalBuilds {
+					directives := ext.RegisterExternalBuilder()
+					for _, d := range directives {
+						extDirectivesMap[d] = ext
+					}
+
+					if ext, ok := extDirectivesMap[directive]; ok {
+						_, _ = sb.WriteString(" ") // space between directives and arguments
+						if err := ext.Build(sb, stmt); err != nil {
+							log.Printf("Failed to write externaller block: %v", err)
```
ornj, line 160: I don't think it's appropriate to log here. If this method returns an error then I think it needs to be returned to the caller, which is not part of this function's signature... Either we need to change the function to return it or design the interface in a way that does not enable returning an error.

lex.go
```diff
@@ -43,14 +43,60 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
 func SetTokenChanCap(size int) {
 	tokChanCap = size
 }
-func Lex(reader io.Reader) chan NgxToken {
+
+// ExternalLexer is an interface for implementing lexers that handle external NGINX tokens during the lexical analysis phase.
+type ExternalLexer interface {
```
ornj, line 48: Up above I suggested removing `External` from the interface name. I think the names look better and more "Go-like" without `External` being part of the interface names, but I see how that feedback can be hard to apply here. We can sync up and figure out if we can think of better names or if we want to just stick with the ones that exist.

build.go
xynicole: Removed the `External`. I have a dumb question for the first question: aren't all directive names unique in NGINX?

lex.go
xynicole: Removed the` External`.

build.go
xynicole: I changed the interface and used the one that we talk in Slack.

build.go
ornj: I cannot offer you a definitive answer for this, but I think this is a good design for a number of reasons. I think of each context as offering a "namespace" without which it would make it even harder to come up with good names for things.

If all directive names were required to be globally unique it might make it harder to implement something like a "virtual server" which is exposed in the configuration as a `server` directive by multiple modules (http, stream, mail...).

build.go
```diff
@@ -131,37 +147,53 @@ func buildBlock(sb io.StringWriter, parent *Directive, block Directives, depth i
...
 			_, _ = sb.WriteString(directive)
 
-			// special handling for if statements
-			if directive == "if" {
-				_, _ = sb.WriteString(" (")
-				for i, arg := range stmt.Args {
-					if i > 0 {
-						_, _ = sb.WriteString(" ")
+			if options.ExternalBuilds != nil {
+				extDirectivesMap := make(map[string]Builder)
```
ornj, line 151: `buildBlock` is called recursively. I don't think you want to rebuild the `extDirectivesMap` map each time. I think it would be better to execute this loop in the `Build` function.

build.go
```diff
@@ -18,9 +18,24 @@ import (
...
+	ExternalBuilds []Builder // handle specific directives
+}
+
+// Builder is the interface implemented by types that can render a Directive
+// as it appears in NGINX configuration files.
+//
+// RegisterBuilder returns the names of the directives for which the builder can
+// build NGINX configuration.
+//
+// Build writes the strings that represent the Directive and it's Block to the
```
ornj, line 33: As implemented, the directive name is written to the string builder first before `Build` is called. Do we want to make that detail here (that the implementation is to return just the arguments and block) or do we want to make the implementation responsible for writing the directive name as well?

lex.go
ornj: I think it's just a minor difference, resolving.

ornj (approved):
I'd like to see docstrings on most of the added exported types/functions where it would help the user understand what to do with them (in case it's not obvious) but other than that, lgtm.

lex.go
```diff
@@ -43,14 +43,95 @@ var tokChanCap = TokenChanCap // capacity of lexer token channel
...
+	return LexWithOptions(reader, LexOptions{})
+}
+
+// SubScanner provides an interface for scanning alternative grammars within NGINX configuration data.
+type SubScanner struct {
+	scanner   *bufio.Scanner
+	tokenLine int
+}
+
+// Scan advances the scanner to the next token which will be availabvle though the Text method. It returns false
```
ornj, line 112:
```suggestion
// Scan advances the scanner to the next token which will be available though the Text method. It returns false
```

lua.go
```diff
@@ -0,0 +1,171 @@
+package crossplane
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Lua struct handles the encapsulation of directives related to Lua module in NGINX configurations.
```
ornj, line 8: Something like:

```suggestion
// Lua adds support for directives added to NGINX by the ngx_http_lua_module module.
// 
// Lua implements the Lexer interface by tokenizing *_by_lua_block directives with a
// simple Lua parser. The Lua blocks will be placed into a token to be set as an 
// argument on the Lua directive.
//
// Lua also implements the Builder interface by writing the the *_by_lua_block 
// directive's contents into the directive's block. 
```

xynicole: Done

lex.go
xynicole: done

lua.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// RegisterBuilder registers a builder for generating Lua NGINX configuration.
+func (l *Lua) RegisterBuilder() RegisterBuilder { //nolint:ireturn
+	return BuildWithBuilder(l, l.directiveNames()...)
+}
+
+// Build generates Lua configurations based on the provided directive.
+func (l *Lua) Build(stmt *Directive) string {
+	if stmt.Directive == "set_by_lua_block" {
```
nginx-nickc, line 173: use const for "set_by_lua_block", since it's in multiple places.

lua.go
```diff
@@ -0,0 +1,178 @@
...
+				if tokenDepth > 0 { // not the last close brace for it to be 0
+					tok.WriteString(next)
+				}
+
+				if tokenDepth == 0 {
+					tokenCh <- NgxToken{Value: tok.String(), Line: s.Line(), IsQuoted: true}
+					tokenCh <- NgxToken{Value: ";", Line: s.Line(), IsQuoted: false} // For an end to the Lua string based on the nginx bahavior
+					// See: https://github.com/nginxinc/crossplane/blob/master/crossplane/ext/lua.py#L122C25-L122C41
+					return
+				}
```
nginx-nickc, line 139: does it make sense to use `switch tokenDepth`?  from line 116 to 139

lua.go
```diff
@@ -0,0 +1,178 @@
...
+
+// RegisterLexer registers a lexer for parsing Lua blocks.
+func (l *Lua) RegisterLexer() RegisterLexer { //nolint:ireturn
+	return LexWithLexer(l, l.directiveNames()...)
+}
+
+// Lex lexically analyzes the Lua blocks based on directives detected.
+// It is used by the lexer to tokenize Lua content within configuration files.
+//
+//nolint:funlen,gocognit,gocyclo,nosec
```
nginx-nickc, line 48: nit, it's totally new code, I'm wondering if we could do better?

lua.go
```diff
@@ -0,0 +1,178 @@
...
+
+				if tokenDepth == 0 {
+					tokenCh <- NgxToken{Value: tok.String(), Line: s.Line(), IsQuoted: true}
+					tokenCh <- NgxToken{Value: ";", Line: s.Line(), IsQuoted: false} // For an end to the Lua string based on the nginx bahavior
+					// See: https://github.com/nginxinc/crossplane/blob/master/crossplane/ext/lua.py#L122C25-L122C41
+					return
+				}
+
+			case next == `"` || next == "'":
+				inQuotes = !inQuotes
```
nginx-nickc, line 142: how do we know the "closing quote" is of the same type? 
if i do the following, wouldn't this mess things up?
```
" ' foo ' "

```

lua.go
```diff
@@ -0,0 +1,178 @@
...
+
+// RegisterBuilder registers a builder for generating Lua NGINX configuration.
+func (l *Lua) RegisterBuilder() RegisterBuilder { //nolint:ireturn
+	return BuildWithBuilder(l, l.directiveNames()...)
+}
+
+// Build generates Lua configurations based on the provided directive.
+func (l *Lua) Build(stmt *Directive) string {
+	if stmt.Directive == "set_by_lua_block" {
+		return fmt.Sprintf("%s %s {%s}", stmt.Directive, stmt.Args[0], stmt.Args[1])
```
nginx-nickc, line 174: length check for args? we shouldn't crash because of invalid input

lua.go
xynicole: I think using `switch tokenDepth `here may not be as effective and readable as the current approach. The current method involves checking conditions based on the value of `next`token to see if we are enter the lua block code

lua.go
xynicole: We are following the Python version of Crossplane https://github.com/nginxinc/crossplane/blob/master/crossplane/ext/lua.py  to implement the Lua part. While there might be some optimizations for this part, considering time constraints, perhaps we could have another PR dedicated to refactoring and optimization.

lua.go
xynicole: I think the analyzer and lexer perform argument length checks, so the number of args passed on `Build` should be correct

lua.go
xynicole: Done

lua.go
xynicole: Good catch, thanks for pointing that out!  I added `var quoteType string` to track if quotes are matched

lua.go
xynicole: updated function now checks if there are enough arguments in the Args

nginx-nickc: nit, can probably just return `stm.Directive`.

nginx-nickc (approved):

lua.go
xynicole: Done, changed return to stm.Directive

PR #88, opened 2024-05-15
Support ngx_mgmt_module

dakshinai:
### Proposed changes

Adding support for `ngx_mgmt_module`

### Checklist

### DIFF
--- a/analyze.go
+++ b/analyze.go
@@ -38,6 +38,7 @@ const (
 
 	// bit masks for different directive locations.
 	ngxDirectConf     = 0x00010000 // main file (not used)
+	ngxMgmtMainConf   = 0x00020000 // mgmt // unique bitmask that may not match NGINX source
 	ngxMainConf       = 0x00040000 // main context
 	ngxEventConf      = 0x00080000 // events
 	ngxMailMainConf   = 0x00100000 // mail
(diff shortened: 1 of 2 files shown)


analyze.go
```diff
@@ -52,10 +52,11 @@ const (
 	ngxHTTPSifConf    = 0x20000000 // http > server > if
 	ngxHTTPLifConf    = 0x40000000 // http > location > if
 	ngxHTTPLmtConf    = 0x80000000 // http > location > limit_except
+	ngxMgmtMainConf   = 0x00000002 // mgmt // TODO: NLB-4310 set correct value here
```
dakshinai, line 55: TBD: to figure out correct value

dakshinai: Set a unique mask w.r.t crossplane

analyze.go
```diff
@@ -1836,6 +1868,9 @@ var directives = map[string][]uint{
 	"upstream_conf": {
 		ngxHTTPLocConf | ngxConfNoArgs,
 	},
+	"usage_report": {
+		ngxMgmtMainConf | ngxConfAny,
```
dakshinai, line 1872: NGINX does accept `usage_report;`

ornj: Then should it be `ngxConfNoArgs | ngxConfTake12`?

analyze.go
```diff
@@ -29,6 +29,7 @@ const (
 	ngxConf2More = 0x00002000 // >=2 args
 
 	// some helpful argument style aliases.
+	ngxConfTake012  = ngxConfNoArgs | ngxConfTake1 | ngxConfTake2
```
ornj, line 32: These mostly match the variables listed in the [Development Guide](https://nginx.org/en/docs/dev/development_guide.html#config_directives) (not values, but variables). We don't export these yet but I think it makes sense to do so soon. I'm not so sure we need to add another "alias" and should just do `ngxConfNoArgs | ngxConfTake1 | ngxConfTake2`  in the directive mapping.

ornj (approved):

xynicole (approved):

PR #95, opened 2024-06-11
Refactor idea to simplify interfaces

ornj:
Moves the "register" responsibility to the options types, not 100% on the API, but it's an idea. The interfaces are now single-method which is pretty much as small as it gets. You could implement these interfaces with a single function that calls itself.

A benefit to this is most obvious with the external Lexer I think as now it is not responsible for storing a value for later use that carries state and is not concurrent-safe.

### Proposed changes

### Checklist


xynicole:
LGTM

xynicole (approved):

PR #97, opened 2024-06-14
Add `DirectiveSources` field to `ParseOptions`

dengsh12:
### Proposed changes

Add a `DirectiveSources` field in `ParseOptions`. This field aims for users to indicate the OSS/N+ version, and dynamic modules they want to include for validation. If it is provided, it will be the only source of directives included in validation, and `ParseOptions.MatchFuncs` will be ignored

### Checklist

### DIFF
--- a/analyze.go
+++ b/analyze.go
@@ -94,13 +94,32 @@ func enterBlockCtx(stmt *Directive, ctx blockCtx) blockCtx {
 
 //nolint:gocyclo,funlen,gocognit
 func analyze(fname string, stmt *Directive, term string, ctx blockCtx, options *ParseOptions) error {
-	masks, knownDirective := directives[stmt.Directive]
-	currCtx, knownContext := contexts[ctx.key()]
+	var masks []uint
+	knownDirective := false
 
-	if !knownDirective {
-		for _, matchFn := range options.MatchFuncs {
-			if masks, knownDirective = matchFn(stmt.Directive); knownDirective {
-				break
+	currCtx, knownContext := contexts[ctx.key()]
+	directiveName := stmt.Directive
+
+	// Find all bitmasks from the sources invoker provides.
+	for _, matchFn := range options.DirectiveSources {
+		if masksInFn, found := matchFn(directiveName); found {
+			masks = append(masks, masksInFn...)
+			knownDirective = true
+		}
+	}
+
+	// If DirectiveSources was not provided, try to find the directive in legacy way.
+	// We want to use DirectiveSources to indicate the OSS/N+ version, and dynamic
+	// modules we want to include for validation. However, invokers used MatchFuncs
+	// and a directive map before. This is a transition plan. After MatchFuncs is deleted
+	// from ParseOptions, this if block should be deleted.
+	if len(options.DirectiveSources) == 0 {
+		masks, knownDirective = directives[directiveName]
+		if !knownDirective {
+			for _, matchFn := range options.MatchFuncs {
+				if masks, knownDirective = matchFn(stmt.Directive); knownDirective {
+					break
+				}
 			}
 		}
 	}
--- a/analyze_test.go
+++ b/analyze_test.go
@@ -2427,3 +2427,125 @@ func TestAnalyze_headers_more(t *testing.T) {
 		})
 	}
 }
+
+//nolint:funlen
+func TestAnalyze_directiveSources(t *testing.T) {
+	t.Parallel()
+	// two self defined maps and matchFn to ensure it is a separate test
+	testDirectiveMap1 := map[string][]uint{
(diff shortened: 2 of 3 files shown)


analyze_test.go
```diff
@@ -2427,3 +2427,88 @@ func TestAnalyze_headers_more(t *testing.T) {
...
+		"only in matchFn ok": {
+			&Directive{
+				Directive: "test_dir",
+				Args:      []string{"args1"},
+				Line:      5,
+			},
+			blockCtx{"http", "upstream"},
+			false,
+		},
+		"in matchFn and OSS but only satisfy matchFn ok": {
```
ornj, line 2458: It's taking me a bit to understand these subtest names. I think "in matchFn and Oss..." means that the directive is matched by both `testMatchFn` and the existing built-in map? I think over time this name might lose meaning as we continue refactoring the way the maps are arranged and directives are "enabled".

When these tests fail for some reason, Go is going to print the test name as something like: `TestAnalyze_matchFn/in_matchFn_and_OSS_but_only_satisfy matchFn_ok` making it a bit harder for people to find the offending subtest as `cmd+F` is not going to work. Consider naming these tests to match the output to make things a little easier.

```suggestion
		"MatchedByFuncAndBuiltinMap": {
```

dengsh12: Sure! Thanks for your suggestion!

dengsh12: I modify "OSS" to "default map" since in the future this will change to a default behavior when we don't pass matchFn for OSS and N+, feel free to provide any further suggestions

ornj: Unresolving because I don't think it's been improved substantially and subtest names still include spaces.

analyze_test.go
```diff
@@ -2427,3 +2427,88 @@ func TestAnalyze_headers_more(t *testing.T) {
...
+		"matched only by matchFn": {
+			&Directive{
+				Directive: "test_dir",
+				Args:      []string{"args1"},
+				Line:      5,
+			},
+			blockCtx{"http", "upstream"},
+			false,
+		},
+		"matched by matchFn and default map case1": {
```
ornj, line 2458: Which one should provide the mask that matches? Is it the built-in map rather than the matchfn or the other way around? how can the test prove this?

dengsh12: case1 is matchFn provides it, default map doesn't and case2 is default map provides, matchFn doesn't. I added some annotations to it. Thanks for pointing it out!

analyze_test.go
dengsh12: I've taken a look at [subtest document](https://go.dev/blog/subtests), add some annotations and change the names, thanks for suggestions.

analyze.go
```diff
@@ -94,13 +94,25 @@ func enterBlockCtx(stmt *Directive, ctx blockCtx) blockCtx {
...
+			}
+		}
+	} else {
+		for _, matchFn := range options.DirectiveSources {
+			if masksInFn, found := matchFn(directiveName); found {
+				masks = append(masks, masksInFn...)
+				knownDirective = true
 			}
 		}
 	}
```
ornj, line 118: If possible, can we simplify this by eliminating some indentation making it easier to read? I'm not sure that the following suggestion works as I haven't tried it and am coding in the comment editor.

```suggestion
	for _, matchFn := range options.DirectiveSources {
		if masksInFn, found := matchFn(directiveName); found {
			masks = append(masks, masksInFn...)
			knownDirective = true
		}
	}
	
	// This could be else if, but then I think you need to document that MatchFuncs
	// will be ignored if DirectiveSources is specified.
	if !knownDirective {
		masks, knownDirective = directives[directiveName]
		for _, matchFn := range options.MatchFuncs {
			if masks, knownDirective = matchFn(stmt.Directive); knownDirective {
				break
			}
		}
	}
```

parse.go
```diff
@@ -108,7 +108,14 @@ type ParseOptions struct {
 	// encountered by the parser to determine the valid contexts and argument count of the
 	// directive. Set this option to enable parsing of directives belonging to non-core or
 	// dynamic NGINX modules that follow the usual grammar rules of an NGINX configuration.
+	// Deprecated: use MatchAny instead
```
ornj, line 111: If the options aren't going to be "validated" to ensure the caller is not mixing the two options, then I think the behavior that these would be ignored should at least be documented.

analyze.go
dengsh12: Cool! This refactor also indicates we prioritize `options.DirectiveSources`. Done in the latest commit

parse.go
dengsh12: Sure! Added it in latest commit

ornj (approved):

nginx-nickc (approved):

PR #98, opened 2024-06-18
Fix lua build

xynicole:
### Proposed changes
Fixed Lua build, updated `buildBlock` logic, and refined unit tests. Previously, we missed cases involving other directives. Our unit tests only checked the Lua directive; now, I’ve added tests for additional directives alongside the Lua directives.

### Checklist

### DIFF
--- a/build_test.go
+++ b/build_test.go
@@ -263,13 +263,49 @@ var buildFixtures = []buildFixture{
 		options: BuildOptions{Builders: []RegisterBuilder{lua.RegisterBuilder()}},
 		parsed: Directives{
 			{
-				Directive: "content_by_lua_block",
-				Line:      1,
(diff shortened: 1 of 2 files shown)


build_test.go
```diff
@@ -263,13 +263,49 @@ var buildFixtures = []buildFixture{
...
+								},
+							},
+						},
+					},
+				},
 			},
 		},
 
-		expected: "content_by_lua_block {\n                ngx.say(\"I need no extra escaping here, for example: \\r\\nblah\")\n            }",
+		expected: strings.Join([]string{
```
ornj, line 298: Alternatively you could have used backticks (``)

xynicole: Gotcha! I was trying to follow the same pattern in this test file

valyria257 (approved):

PR #100, opened 2024-06-26
Provides more `MatchFunc` and code movement for it

dengsh12:
### Proposed changes

In #97 , we provided a new field `DirectiveSources` in `ParseOptions`. It is an array of `MatchFunc` to specify the OSS/N+ version, and dynamic modules to include for validation. 

In this PR, we
- Provide enough `MatchFunc` for OSS/N+ version, and dynamic modules, which can be passed to `ParseOptions .DirectiveSources`.
- Delete `ParseOptions.MatchFuncs`, users should use `ParseOptions.DirectiveSources` instead.
- Modify some logics to accommodate it.
- Move some codes from one file to separate files for better management

### Checklist


analyze.go
```diff
@@ -108,19 +108,17 @@ func analyze(fname string, stmt *Directive, term string, ctx blockCtx, options *
 		}
 	}
 
-	// If DirectiveSources was not provided, try to find the directive in legacy way.
-	// We want to use DirectiveSources to indicate the OSS/N+ version, and dynamic
-	// modules we want to include for validation. However, invokers used MatchFuncs
-	// and a directive map before. This is a transition plan. After MatchFuncs is deleted
-	// from ParseOptions, this if block should be deleted.
+	// If DirectiveSources was not provided, use the default map
```
ornj, line 111: "use the default map" sounds like a direction to the reader. What I think you mean to communicate is that if the `DirectiveSources` option is not provided (is empty?) then the parser will match against a default set of directives.

analyze.go
```diff
@@ -108,19 +108,17 @@ func analyze(fname string, stmt *Directive, term string, ctx blockCtx, options *
...
-			for _, matchFn := range options.MatchFuncs {
-				if masks, knownDirective = matchFn(stmt.Directive); knownDirective {
-					break
-				}
-			}
+	}
+
+	if knownDirective {
+		// Check if the directive has human-defined special definition
+		// If so, use it to do validation
```
ornj, line 118: I found the comment to be a bit oddly worded, but I think you shouldn't need it anyway. I think if `forcedMap` were named better then the code could more clearly communicate what is happening here.

analyze.go
```diff
@@ -108,19 +108,17 @@ func analyze(fname string, stmt *Directive, term string, ctx blockCtx, options *
...
+		bitmaskInForceMap, found := forcedMap[directiveName]
+		if found {
+			masks = bitmaskInForceMap
 		}
```
ornj, line 122: Nit

```suggestion
		if m, ok := forcedMap[directiveName]; ok {
			masks = m
		}
```

forced_directives_map.go
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) F5, Inc.
+ *
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package crossplane
+
+// A human editable map. Used to overwrite generated directive definitions for special cases.
```
ornj, line 10: It's unclear to me what "forcedMap" means. I think there might be a better name for this variable. I also think that the comment isn't adding clarity. I'm not sure why `if` is a special case necessitating it be put into it's own map. 

All code is by default "editable". It's a common Go convention to add a comment at the top of the file containing generated code that should not be edited directly, something like:

```go
// Code generated by thinger -option Thingy thing.go; DO NOT EDIT.
```

ornj:
I think all the code with the directive maps should have a `analyze_` prefix instead of `module_` or anything else. This way they will appear in editors near the main `analyze.go` file.

ornj (changes requested):

forced_directives_map.go
dengsh12: > I'm not sure why `if` is a special case necessitating it be put into its own map.

The reason is: `if` has a  "crossplane created" bitmask `ngxConfExpr`. `ngxConfExpr` was not in source code and crossplane created it to indicate a directive followed by expression in parentheses `()`. I may caution there will be similar cases in the future so I want to provide a way for human to override the generated map

> It's a common Go convention to add a comment at the top of the file containing generated code that should not be edited directly, something like:

I had `// Code generated by generator; DO NOT EDIT.` in these files but not at the top. I think copyright should appear first?(I failed to find an example, chatGPT said copyright first)

> I think there might be a better name for this variable

Sure.... I struggled a lot for it. Just thought a name `priorityDirectiveMap`. Maybe it's better?

priority_directives_map.go
```diff
@@ -0,0 +1,19 @@
...
+// A map used to overwrite generated directive definitions.
+// When we have a human-defined definition for a directive, which is different
+// from the definition in source code, put it here.
+//
+//nolint:gochecknoglobals
+var priorityMap = map[string][]uint{
+	"if": {
+		ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfBlock | ngxConfExpr | ngxConf1More,
+	},
+}
```
ornj, line 19: How about something like this. I've renamed the variable to match it's intended use and included detail from your [comment](https://github.com/nginxinc/nginx-go-crossplane/pull/100#discussion_r1655444060) about the `if` directive.

```go
// directiveOverrides is a map of directive names to masks used to override any
// masks that are provided to tell the parser how to validate directives and their
// arguments.
//nolint:gochecknoglobals
var directiveOverrides = map[string][]uint{
	// if contains a bitmask (ngxConfExpr) that does not exist in NGINX that indicates
	// the directive name is followed by an expression in parantheses.
	"if": {
		ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfBlock | ngxConfExpr | ngxConf1More,
	},
}
```

ornj: I'm wondering if this has to obliterate any masks passed by the caller, or if this should instead be appended to the list of masks? It may be unlikely, but what if another module provides an `if` directive to be used in other contexts?

dengsh12: > How about something like this. I've renamed the variable to match it's intended use and included detail from your https://github.com/nginxinc/nginx-go-crossplane/pull/100#discussion_r1655444060 about the if directive.

Sure! It seems more clear. And the comments for the reason we have `if` here makes it much better.

> I'm wondering if this has to obliterate any masks passed by the caller, or if this should instead be appended to the list of masks?

If we apply "append to DirectiveSources", the `if` statement can pass the validation even if no `()` after it. The reason is current validation logic for bitmasks is `or`. In "append to DirectiveSources", `if` can pass validation through the bitmask from default map(or `DirectiveSources`). By overriding user has high flexibility, if in the future a new bitmask of `if` appears in OSS, user can add the new bitmask to `directiveOverrides`.

I've thought of providing a "overrideMap" and a "appendMap" before, but it seems the "appendMap" is unnecessary since user can put all bitmasks in overrideMap to achieve its function.

> It may be unlikely, but what if another module provides an if directive to be used in other contexts?

I think this situation it's kind of out of current scope? To address this we need to define a "override map" for every module, when the module is included, the corresponding "override map"  should be included together. But it seems very unlikely to happen since we only have a case of `if` till now.

ornj: Say I implement my own `ngx_foo.so` module that adds a new context and with a new directive called `if` that implements a different syntax.

```nginx
foo {
    if $arg_foo == "";
}
```

How can I as a module author provide a `MatchFunc` to handle my new directive that occurs in the `foo` context? I am less concerned about `if` specifically and more concerned about future expansion.

parse.go
```diff
@@ -104,17 +104,12 @@ type ParseOptions struct {
...
-	// If DirectiveSources was provided, this will be ignored.
-	MatchFuncs []MatchFunc
-
 	// An array of MatchFunc, which is used to indicate the OSS/N+ version, and
 	// dynamic modules you want to include for validation. If a directive matches
 	// any of them, and satisfies the corresponding bitmask, it should pass the validation.
-	// If we provide this, MatchFuncs will be ignored
+	// If this is provided, it will be the only source of directives included for validation.
+	// If this is not provided, `defaultDirectives`(NPlus, Njs, and Otel in it) will be included
+	// for validation.
```
ornj, line 112: Generally doc comments should generally have the exported name in the first line. I think this is a generally useful convention as it results in the first line also appear when searching for an exported member providing a little context:

```
$ grep "MatchFunc" parse.go
// MatchFunc is the signature of the match function used to identify NGINX directives that
type MatchFunc func(directive string) (masks []uint, matched bool)
	// MatchFuncs are called in order when an unknown or non-core NGINX directive is
	MatchFuncs []MatchFunc
```

Something like:

```go
	// DirectiveSources is a list of MatchFuncs used to validate directives 
	// encountered while parsing an NGINX configuration... 
```

The current comment mentions `defaultDirectives`-- probably you want to omit the backticks from the comment-- I think you want to mention the exported `DefaultDirectivesMatchFunc` function instead as the unexported variable is not useable from outside the package.

parse.go
```diff
@@ -104,17 +104,12 @@ type ParseOptions struct {
...
-	MatchFuncs []MatchFunc
-
-	// An array of MatchFunc, which is used to indicate the OSS/N+ version, and
+	// DirectiveSources is an array of MatchFunc, which is used to indicate the OSS/N+ version, and
 	// dynamic modules you want to include for validation. If a directive matches
 	// any of them, and satisfies the corresponding bitmask, it should pass the validation.
-	// If we provide this, MatchFuncs will be ignored
+	// If this is provided, it will be the only source of directives included for validation.
+	// If this is not provided, DefaultDirectivesMatchFunc(NPlus, Njs, and Otel in it) will be
+	// used for validation.
```
ornj, line 112: Nit: `DirectiveSources` is a `slice` not an `array`. I think "list" is also an acceptable, inexact noun to use. I think we can tidy up the language a bit:

```go
// DirectiveSources is used to indicate the set of directives to be expected
// by the parser. Directive sources can include different version of NGINX
// and dynamic modules. If DirectiveSources is empty, the parser defaults
// to DefaultDirectivesMatchFunc.
```

The doc comment for `DefaultDirectivesMatchFunc` how that option behaves.

ornj (approved):

parse.go
dengsh12: Good suggestion! Fixed

dakshinai:
@ornj @dengsh12 What will be the process to add a new N+ version, say R32, add it to NPlusLatest?

I do see see an overlapping list for R30 and R31, I believe specific version lists should include everything up until the version to be able to perform version specific validation? So for R32, create one if needed?

dengsh12:
> @ornj @dengsh12 What will be the process to add a new N+ version, say R32, add it to NPlusLatest?

@dakshinai It will be in later merge. I will provide a generator. If a new N+ version comes just run `go generate ./...`. It will fetch from some sources and create new support files.

> I do see see an overlapping list for R30 and R31, I believe specific version lists should include everything up until the version to be able to perform version specific validation?

The support file for a specific version only includes the directive definitions in the source code of that version. Say the version is R31, `analyze_nplus_R31_directives.go` only includes the directive definitions in Nplus R31 source code(so if a directive was removed before R31 it won't appear). The parameter `DirectiveSources` provides high flexibility, you can pass `MatchFunc` for both R30 and R31 to include both of them for validation. I think this way is more flexible and clear?

>  So for R32, create one if needed?

Currently latest version is R32 so just use `analyze_nplus_latest_directives.go` can cover. This file will update, say R33 will release tomorrow and this file will update to R33 tomorrow, and a `analyze_nplus_R32_directives.go` will be generated(since it is not latest). I may caution provide both R32 and `analyze_nplus_latest_directives.go` will make it confused?

Feel free to lmk if there're further concerns!

analyze_nplus_R30_directives.go
```diff
@@ -0,0 +1,2112 @@
...
+// Each bit mask describes these behaviors:
+//   - how many arguments the directive can take
+//   - whether or not it is a block directive
+//   - whether this is a flag (takes one argument that's either "on" or "off")
+//   - which contexts it's allowed to be in
+
+package crossplane
+
+//nolint:gochecknoglobals
+var ngxPlusR30Directives = map[string][]uint{
```
dakshinai, line 18: from the code, 
NgxPlusR31DirectivesMatchFn has the list of all directives supported until R31 and not purely only those introduced with R31? So,

To validate my config against a cutoff version - support everything <=R30
DirectiveSources:         []MatchFunc{NgxPlusR30DirectivesMatchFn}

To validate my config against a cutoff version - support everything <=R31
DirectiveSources:         []MatchFunc{NgxPlusR31DirectivesMatchFn}

If there are overlaps, we would pass the directive based on its first match, we ignore the order?
DirectiveSources:         []MatchFunc{NgxPlusR30DirectivesMatchFn, NgxPlusR31DirectivesMatchFn}

dengsh12: > NgxPlusR31DirectivesMatchFn has the list of all directives supported until R31 and not purely only those introduced with R31?

Nah, it only has directives in R31 source code, not all directives supported until R31. If a directive appears in R31 source code it should appear in `NgxPlusR31Directives`, otherwise not.

> To validate my config against a cutoff version - support everything <=R30

For this case we should go in this way: DirectiveSources: []MatchFunc{NgxPlusR30DirectivesMatchFn, NgxPlusR29DirectivesMatchFn, NgxPlusR28DirectivesMatchFn,...}. I assumed we only need the latest 3 versions, but if needed I can add MatchFn for R29, R28, or more.

> If there are overlaps, we would pass the directive based on its first match, we ignore the order?
DirectiveSources: []MatchFunc{NgxPlusR30DirectivesMatchFn, NgxPlusR31DirectivesMatchFn}

We ignore the order, get all bitmasks from all DirectiveSources. If a directive aligns with any bitmask from any of them, it can pass the validation. For example, say `absolute_redirect` has bitmask1,bitmask2 in NgxPlusR30DirectivesMatchFn, bitmask3 in NgxPlusR31DirectivesMatchFn. It can pass the validation if it satisfy bitmask1, bitmask2, or bitmask3

dakshinai: `analyze_nplus_R30_directives.go` has all directives up until R30 and hence I inferred
To validate my config against a cutoff version - support everything <=R30
`DirectiveSources: []MatchFunc{NgxPlusR30DirectivesMatchFn}`

dakshinai (approved):

analyze_lua_directives.go
```diff
@@ -0,0 +1,261 @@
...
+	"lua_capture_error_log": {
+		ngxHTTPMainConf | ngxConfTake1,
+	},
+	"lua_check_client_abort": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfFlag,
+	},
+	"lua_code_cache": {
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfFlag,
+	},
+	"lua_fake_shm": {
```
xynicole, line 106: This directive is not introduced in the Lua docs https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#directives

dengsh12: It is a directive in [/lua-nginx-module/t/data/fake-shm-module/ngx_http_lua_fake_shm_module.c](https://github.com/openresty/lua-nginx-module/blob/master/t/data/fake-shm-module/ngx_http_lua_fake_shm_module.c). I guess they want to drop the support after a period so it is undocumented

xynicole: https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_module.c 
we reference to this file

dengsh12: Both of these two files are in the [lua git repo](https://github.com/openresty/lua-nginx-module/tree/master). And it seems there is still a [test case](https://github.com/openresty/lua-nginx-module/blob/master/t/149-hup-fake-shm-zone.t) for `lua_fake_shm`. The last commit for `lua_fake_shm ` is 9 years ago. I think this directive is likely an outdated Lua directive(not pretty sure)?

I just run a script and it seems [ngx_http_lua_module.c](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_module.c) covers all directives except `lua_fake_shm`. Currently the generator can rely on it, but I'm afraid they may split `ngx_http_lua_module.c` or have some directive elsewhere. What about generate for all directives on [lua git repo](https://github.com/openresty/lua-nginx-module/tree/master), then use the [lua doc](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#directives) as a filter? I can add it in next MR

PR #101, opened 2024-07-03
Command line code generator for supporting files

dengsh12:
### Proposed changes

In the past, when people want to add support for new dynamic modules, or update current existing directive definitions in nginx-go-crossplane, they always need to look at documents or source codes and find the definition of directives.

In this PR, we provide a command line tool, which can read from a directory containing source codes, and generate the corresponding supporting file. The supporting file will include a map from directives to their bitmask definitions, and a public `MatchFunc` to match them. It will be in the same format of currently existing supporting files like `analyze_otel_directives.go`, `analyze_oss_latest_directives.go`.

### Checklist


internal/generator/generator.go
```diff
@@ -0,0 +1,16 @@
...
+
+package generator
+
+import (
+	"os"
+)
+
+func Generate(sourcePath string) error {
+	return genSupFromSrcCode(sourcePath, "directives", "Match", os.Stdout)
+}
```
ornj, line 16: I think this would be more useful if implemented like this,

```go
// Generate... fill in missing doc comment
func Generate(w io.Writer, path string) error
```

This way any `io.Writer` can be passed. `cmd/generate/main.go` could pass `os.Stdout` and you can pass `bytes.Buffer` in tests. You could even pass an `http.ResponseWriter` if you wanted to return the result to an http request for some reason.

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+	// Extract one directive definition and attributes from extracted block
+	// { ngx_string({directive_name}),
+	//   {bitmask1|bitmask2|...},
+	//   ... },
+	// this regex extracts {directive_name} and {bitmask1|bitmask2|...}.
+	extractSingleDirective = "ngx_string\\(\"(.*?)\"\\).*?,(.*?),"
+
+	extractSingleLineComment = `//.*`
+
+	extractMultiLineComment = `/\*[\s\S]*?\*/`
```
ornj, line 39: Nit: Unless you need to reference these strings in a few different places, I think it's better to move this down the the call to `regexp.MustCompile`

```go
var (
	directivesDefBlockExtracter = regexp.MustCompile("ngx_command_t\\s+(\\w+)\\[\\]\\s*=\\s*{(.*?)};")
	// ...rest
)
```

De-duplicating string constants can be a good idea, but I think when applied too aggressively it hurts readability.

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+	// static ngx_command_t  {name}[] = {definition}
+	// this regex extracts {name} and {definition}.
+	extractDirectivesDefBlock = "ngx_command_t\\s+(\\w+)\\[\\]\\s*=\\s*{(.*?)};"
+
+	// Extract one directive definition and attributes from extracted block
+	// { ngx_string({directive_name}),
+	//   {bitmask1|bitmask2|...},
+	//   ... },
+	// this regex extracts {directive_name} and {bitmask1|bitmask2|...}.
+	extractSingleDirective = "ngx_string\\(\"(.*?)\"\\).*?,(.*?),"
```
ornj, line 35: Can you explain what these regex patterns do differently from the regex patterns used by @dareste in his similar tool?

```
ngx_command_t(.*?)\[](?s)(.*?)ngx_null_command
ngx_string\("(.*?)"\),(.*?),
```

You can use backticks to avoid needing to escape `"` characters.

dengsh12: > Can you explain what these regex patterns do differently from the regex patterns used by @dareste in his similar tool?

His `ngx_command_t(.*?)\[](?s)(.*?)ngx_null_command` and my `extractDirectivesDefBlock` have the same function. But I'm afraid some authors may not put ngx_null_command at the end of definition(still align with C grammar). So I choose the way of `extractDirectivesDefBlock` (I think "};" must be the end of the definition according to C grammar).

His `ngx_string\("(.*?)"\),(.*?)` and my `extractSingleDirective` are same but I did't use backticks, I will change to backticks

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+	"ngxStreamUpsConf":  nil,
+	"ngxHTTPMainConf":   nil,
+	"ngxHTTPSrvConf":    nil,
+	"ngxHTTPLocConf":    nil,
+	"ngxHTTPUpsConf":    nil,
+	"ngxHTTPSifConf":    nil,
+	"ngxHTTPLifConf":    nil,
+	"ngxHTTPLmtConf":    nil,
+	"ngxMgmtMainConf":   nil,
+}
```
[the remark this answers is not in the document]
ornj: Have you come across any examples of dynamic modules creating their own contexts? Without spending too much time looking I've only found examples that work something like `mgmt` (`geoip2`, `otel`).

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+		// The name of the directives block in source code, it may be used as the context
+		blockName := block[1]
+		// Extract directives and their attributes in the code block, the first dimension of attributesList
+		// is index of directive, the second dimension is index of attributes
+		attributesList := singleDirectiveExtracter.FindAllStringSubmatch(block[2], -1)
+		// Iterate through every directive
+		for _, attributes := range attributesList {
+			// Extract attributes from the directive
+			directiveName := strings.TrimSpace(attributes[1])
+			diretiveBitmasks := strings.Split(attributes[2], "|")
```
ornj, line 150:
```suggestion
			directiveBitmasks := strings.Split(attributes[2], "|")
```

internal/generator/generator_util.go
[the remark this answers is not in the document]
dengsh12: > The better way to do this is with a map[string]struct{}.

Sure! Thanks for telling me this. Will fix in next commit

> I've also seen this as map[string]bool that I think is worse because there are 3 possible values: unset, true, or false.

I found this only in an existing test `TestAnalyze`. It seems nothing to do with this MR, maybe I can change it in another separate MR?

> Have you come across any examples of dynamic modules creating their own contexts?

No at all... The only new context I've met is `ngxMgmtMainConf`(and this is created by us, not in source code). I have a map from C style context to go style and if there is a new context not in map I will get an error. Till now I haven't met an error for it.

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+	directive2Defs := make(map[string][]bitDef, 0)
+
+	err := filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		// Check if the entry is a C/C++ file
+		// Some dynamic modules are written in C++, like otel
+		if !d.IsDir() && (strings.HasSuffix(path, ".c") || strings.HasSuffix(path, ".cpp")) {
```
ornj, line 195: Using early returns would help you eliminate some nested and make the code a little easier to read. For example:

```go
if d.IsDir() {
	return nil
}

if !strings.HasSuffix(path, ".c") && strings.HasSuffix(path, ".cpp") {
	return nil
}

// "happy" path is now aligned left. Continue handling the entry.
```

internal/generator/generator_util.go
```diff
@@ -0,0 +1,238 @@
...
+			dir2defsInFile, err := getDirectiveDefFromFile(path)
+			if err != nil {
+				return err
+			}
+			for directive, defsInFile := range dir2defsInFile {
+				if _, found := directive2Defs[directive]; !found {
+					directive2Defs[directive] = []bitDef{}
+				}
```
ornj, line 203: Not necessary. You can `append()` a nil slice.

internal/generator/generator.go
dengsh12: Fixed: applied `func Generate(w io.Writer, path string) error`

internal/generator/generator_util.go
dengsh12: fixed: renamed

internal/generator/generator_util.go
dengsh12: Cool! Beautiful now. Fixed

internal/generator/generator_util.go
dengsh12: Amazing... applied

internal/generator/generator_util_test.go
```diff
@@ -0,0 +1,147 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			res, err := validateTestOut(buf.Bytes(), expectedFile)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if res == false {
+				t.Fatal("output not align with expectation")
```
ornj, line 149: Printing the difference would aid debugging

internal/generator/generator_util.go
```diff
@@ -0,0 +1,235 @@
...
+	"ngxHTTPUpsConf":    {},
+	"ngxHTTPSifConf":    {},
+	"ngxHTTPLifConf":    {},
+	"ngxHTTPLmtConf":    {},
+	"ngxMgmtMainConf":   {},
+}
+
+//nolint:gochecknoglobals
+var directiveBlock2Context = map[string]string{
+	"ngx_mgmt_block_commands": "ngxMgmtMainConf",
```
ornj, line 115: This is the name of the array of commands defined by the `mgmt` module? What should happen when there is another module that handles context in the same way? What happens if the module implements more than one context?

Maybe this is OK for now as an acceptable limitation.

dengsh12: > What should happen when there is another module that handles context in the same way

Say there is a module `MD`, have an array of directive `ngx_MD_block_commands`, all directives here have no context, but actually they should be in block `MD{}` In this case we should add `"ngx_MD_block_commands": MDContext ` into `directiveBlock2Context`, and add `MDContext` into crossplane. I think since `MDGoContext ` is a new context and not in source code we must update crossplane manually for it? 

> What happens if the module implements more than one context?

The assumption I made is: `ngx_mgmt_block_commands` ends with `block_commands` and the "no context directives" in it should actually have a context of `ngx_mgmt_block`(I assume no 2 context in same `block_commands`). If there is another context in this situation we can just add one item into `directiveBlock2Context`?

dareste: > What happens if the module implements more than one context?

With my approach, I can deal with more than one new context per module as long as we make certain assumptions in the order in which the block and directive arrays are defined. However, that's not a robust approach either. I think any silver bullet code might help until we see more examples in future releases.

internal/generator/generator_util_test.go
dengsh12: Fixed:
```
t.Fatalf(
"output not align with expectation, output:\n %s \nexpected:\n %s ",
string(buf.String()), string(expected))
```

ornj: I think the differences might be difficult to spot. Some times the test output is large.

Consider using `testify/assert` or `testify/require` as it will print the diff.

internal/generator/generator_util.go
ornj: I think the ultimate solution for flexibility and fewer assumptions may be to pass some value in to allow some custom behavior? Something like:

```sh
generate -context ngxMgmtMainConf  # add to all discovered directives
generate -context usage_report=ngxMgmtMainConf. # add to specific directive
```

It could get long if there are many directives that need to be touched. I suppose you could solve that somehow with some sort of regex selector syntax? Not sure that we would need it yet.

dengsh12: > Something like:
generate -context ngxMgmtMainConf  # add to all discovered directives
generate -context usage_report=ngxMgmtMainConf. # add to specific directive

I'm not sure if it should be an option. My caution is when user doesn't pass the `-context` flag, the generated "no mgmt context but is mgmt block directives" seems a bit weird.

> It could get long if there are many directives that need to be touched. I suppose you could solve that somehow with some sort of regex selector syntax?

To the top of my head the pattern for these ngxMgmtMainConf directives is: all directives in `ngx_mgmt_block_commands` should be added a ngxMgmtMainConf. And then it turns back to the assumption of those ending with `block_commands`...

internal/generator/generator_util_test.go
dengsh12: Wow, cool! Fixed

internal/generator/generator_util.go
ornj: I think we probably don't need to expose more API to let the caller of the command configure it any further at this time. It was for sake of argument. I do think the assumption of the way commands are defined and having to hard code that is a little suspect.

When I look at @dareste's solution I think it doesn't make that same assumption, but it does assume a certain order to commands. This might also work for other packages that do not define a context via flag (I think `otel` and `geoip2` do this). Perhaps his solution of capturing the "last command block" is a bit more flexible, although I'm unsure the best way to translate that directive's name into a flag.

dengsh12: > When I look at @dareste's solution I think it doesn't make that same assumption

Maybe I made some confusion here lol. I meant I was inspired by his but not exactly same way. I choose this way because at that time I only knew the example of `mgmt` and I thought the pattern "xxx_ block_commands" is reliable.

> This might also work for other packages that do not define a context via flag (I think otel and geoip2 do this).

I see. I didn't notice the cases of `otel` and `geoip2` before. It seems [geoip2](https://github.com/leev/ngx_http_geoip2_module) directives are in `ngxHTTPMainConf`, but otel has some no-context directives in [gExporterCommands](https://github.com/nginxinc/nginx-otel/blob/main/src/http_module.cpp). It seems my former assumption that it should end with `block_commands` doesn't work, but the name of the array  `gExporterCommands` still relate to the context `otel_exporter`. If we want a new context for `otel_exporter` we can add the mapping from `gExporterCommands` to the new context.

I think the problem boils down to depend on "last command block" or "array name". I found both can work currently but both can fail in the future... Since the points we rely on are not "hard rule".

> Perhaps his solution of capturing the "last command block" is a bit more flexible

Could you specify a bit for it? I'm unsure about the flexibilities and appreciate any insights.

dareste: @dengsh12 `otel` is a good example of new context creation as in the `mgmt` case. 

I think you're right, tracking the array of no-context commands and trying to relate it with existing commands with valid contexts might be more robust than the current regex approaches.

This logic could work:

- Find command arrays.
- Loop the commands in the command array.
- IF the command in the array has both `NGX_CONF_BLOCK` and `NGX_HTTP_*_CONF` flags, and also has a non-standard set handler function (see [Module Config Directives](https://nginx.org/en/docs/dev/development_guide.html#config_directives) for a list of predefined handlers)
  - THEN `findRelatedCommands`

`findRelatedCommands` would contain a non-trivial logic to inspect the found handler function ([`setExporter`](https://github.com/nginxinc/nginx-otel/blob/d3817136395a326dd9626580d81e0c4703fc1463/src/http_module.cpp#L591) as the handler of `otel_exporter`, for instance), explore it and find usage references to command arrays without context flag ([`gExporterCommands`](https://github.com/nginxinc/nginx-otel/blob/main/src/http_module.cpp#L604), following the example). All those found commands would belong to a new context defined by the command name with the custom handler (`otel_exporter`).

dengsh12: Cool! `findRelatedCommands` seems the correct way. Searching into the module self defined config handler and figure the relationship should be enough. 
@ornj I'm unsure in crossplane we need to make it such heavy. From another aspect, what we're doing here is to provide a "hidden override". Since we will provide `-filter` and `-override` in next MR, could we just defer the non-context cases into next MR and provide some "module recommended override" to address this? We need to maintain the  "module recommended override", but if a new non-context directive comes we may aslo need to add a new context(even though we have `findRelatedCommands` we may need, since the new context may not be a variable in crossplane )

dengsh12: > but if we need to modify the generator code each time we need to support a new module then I think that degrades the value of having the generator a bit.

Nah, only when the module define a context in such a hidden way we need to modify. If we want to add context for it, we must modify the crossplane since the context is new to crossplane. I wonder if we can just utilize the "-override" option in next MR, since we are basically automatically override it.

For example, if a new module `M` defines a no-context directive `m_non_context`, and a block directive `m_block`. The "actual" context of `m_non_context` is `m_block`, in such situation we can't avoid the effort to add a context for `m_block`. The problem boils down to use some mechanism to scan the `M` source code and  figure out the "actual context" in its self-define config handler(it depends on our assumption, the config handler maybe unpredicted but currently reliable), or just utilize the "-override" to handle `m_non_context`.

ornj: > I wonder if we can just utilize the "-override" option in next MR, since we are basically automatically override it.

👍 I think a later pull-request is fine.

ornj: I don't think I have anything more. If there's no further feedback from @dareste I think it's ok to resolve.

internal/generator/generator_util_test.go
```diff
@@ -0,0 +1,166 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if *update {
+				_, err = expectedFile.WriteString(buf.String())
+				if err != nil {
+					t.Fatal(err)
+				}
+				expectedFile.Close()
```
ornj, line 155: You should not ignore the error returned by `Close` when you are writing the file. This error may indicate that the write has failed. Ignoring the error when you are only reading is generally OK.

ornj: There's another potential bug here where `expectedFile` is only closed if `*update == true` and `WriteString` did not return an error. While I don't think it will ever cause an issue and everything will be cleaned up when the test run exits, it's good practice to make sure you aren't leaking file descriptors. 

```go
expectedFile, err := os.OpenFile(expectedFilePth, os.O_RDWR|os.O_CREATE, 0644)
if err != nil {
	t.Fatal(err)
}

defer func() {
	if err := expectedFile.Close(); err != nil {
		t.Fatal(err)
	}
}()

if *update {
	_, err = expectedFile.Write(buf.Bytes()).
	if err != nil {
		t.Fatal(err)
	}
	return
}
```

internal/generator/generator_util_test.go
```diff
@@ -0,0 +1,170 @@
...
+var (
+	update = flag.Bool("update", false,
+		`update the expected output of these tests, 
+only use when the expected output is outdated and you are sure your output is correct`)
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	os.Exit(m.Run())
+}
```
ornj, line 70: Nit: I think it would be nice to move this to the top of the file so it's easier to spot.

internal/generator/generator_util_test.go
dengsh12: > You should not ignore the error returned by Close when you are writing the file.

Oh I see. Thanks for pointing it out!

> While I don't think it will ever cause an issue and everything will be cleaned up when the test run exits, it's good practice to make sure you aren't leaking file descriptors.

Sure! Committed

internal/generator/generator_util_test.go
dengsh12: Sure! Committed

internal/generator/generator_util.go
dareste: Looks good to me!

cmd/generate/main.go
```diff
@@ -0,0 +1,28 @@
...
+	var (
+		sourceCodePath = flag.String("src-path", "",
+			"the path of source code your want to generate support from, it can be either a file or a directory. (required)")
+	)
+	flag.Parse()
+	err := generator.Generate(*sourceCodePath, os.Stdout)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
```
valyria257, line 28: Can you update the readme to include an example of using this new command line tool?

internal/generator/generator_util.go
```diff
@@ -0,0 +1,200 @@
...
+func genFromSrcCode(codePath string, mapVariableName string, matchFnName string, writer io.Writer) error {
+	directive2Masks, err := getMasksFromPath(codePath)
+	if err != nil {
+		return err
+	}
+
+	err = supportFileTmpl.Execute(writer, supportFileTmplStruct{
+		Directive2Masks: directive2Masks,
+		MapVariableName: mapVariableName,
+		MatchFnName:     matchFnName,
```
valyria257, line 193: Are `MapVariableName` and `MatchFnName` expected to change? Just curious why you made them template variables

dengsh12: Yeah there will be another generator directly fetch code from some github repos and these names will change accordingly. For example, the generator fetches Lua source code and the generated `MapVariableName` will be `moduleLuaDirectives` and the `MatchFnName` will be `LuaDirectivesMatchFn`. That tool aims for auto update current existing support files(`analyze_xxx_directives.go`)

valyria257: Got it, thanks for the explanation!

valyria257 (approved):

internal/generator/generator_util.go
```diff
@@ -0,0 +1,200 @@
...
+	"NGX_DIRECT_CONF":      "ngxDirectConf",
+	"NGX_CONF_TAKE13":      "ngxConfTake13",
+	"NGX_CONF_ANY":         "ngxConfAny",
+	"NGX_CONF_TAKE4":       "ngxConfTake4",
+}
+
+//nolint:nonamedreturns
+func masksFromFile(path string) (directive2Masks map[string][]mask, err error) {
+	directive2Masks = make(map[string][]mask, 0)
+	byteContent, err := os.ReadFile(path)
```
nginx-nickc, line 101: i'm wondering if we want to follow existing convention, by allowing reading of file from a FS (https://github.com/nginxinc/nginx-go-crossplane/blob/main/parse.go#L77), and by default open from the os. https://pkg.go.dev/os#DirFS. This would have the flexibility of using the embed for testing (https://pkg.go.dev/embed), instead of relying on reading the file system.

cmd/generate/main.go
dengsh12: Updated readme, lmk if you have further concerns!

internal/generator/generator_util.go
dengsh12: Let's make sure first, do you mean test in this way?
```go
//go:embed testdata/*
var testFiles embed.FS
for _, test := range tests {
    readEmbed(test.path)
}
```
I thought currently user cases of the generator are to read from files and generate, and for unit tests I thought may not need to add files into compilation? Since these files are only used for one time in unit test, put them in compile won't be much faster then read using `os.Open`. I may caution adding to many unnecessary files(maybe some large file in future) into compilation. Maybe we can do that after the need of reading from sources other than local file system comes, or some ut reads files repeatedly and takes too much time?

PR #102, opened 2024-07-04
fix: Memory leak caused by openFile and osOpen functions

oliveromahony:
### Proposed changes

The parse function was leaking memory. On analysis of the code it looked like the fileOpen function was keeping the memory active, after the file was being used. 

* Changed the io.Reader to io.ReadCloser in Open signature
* Changed the io.Reader to io.ReadCloser in osOpen signature
* Added defer to close the file after function exit in Parse

A snapshot of the memory in the NGINX Agent before these changes:
image

and after:
image

### Checklist

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -32,7 +32,6 @@ func (fs *FileString) Write(b []byte) (int, error) {
 // Close makes this an io.Closer.
 func (fs *FileString) Close() error {
 	fs.w.WriteByte('\n')
-	fs.w.Reset()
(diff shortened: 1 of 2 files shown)


dhurley (approved):

ornj (approved):
Can you also remove this [type assertion block](https://github.com/nginxinc/nginx-go-crossplane/blob/main/parse.go#L399)? I think it is no longer needed now that the correct type is passed.

PR #103, opened 2024-07-08
Add lua analyzer to stream and upstream

rikatz:
### Proposed changes
Some directives are missing on NGINX stream, and some other inside main HTTP server.

This PR fixes the missing directives

### Checklist

### DIFF
--- a/analyze_lua_directives.go
+++ b/analyze_lua_directives.go
@@ -29,7 +29,7 @@ var moduleLuaDirectives = map[string][]uint{
 		ngxHTTPMainConf | ngxConfFlag,
 	},
 	"balancer_by_lua_block": {
-		ngxHTTPUpsConf | ngxConfTake1,
+		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPUpsConf | ngxStreamMainConf | ngxStreamSrvConf | ngxStreamUpsConf | ngxConfTake1,
 	},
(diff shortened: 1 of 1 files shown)


ornj:
Thank you for submitting a PR @rikatz.

The current set of Lua directives supported by crossplane focus on supporting [ngx_http_lua_module](https://github.com/openresty/lua-nginx-module). I think the changes you're making are releated to [ngx_stream_lua_module](https://github.com/openresty/stream-lua-nginx-module).

@xynicole @dengsh12 Should we support this module by merging the flags into `moduleLuaDirectives` or do we want to model it as a separate optional module?

dengsh12:
> Thank you for submitting a PR @rikatz.
> 
> The current set of Lua directives supported by crossplane focus on supporting [ngx_http_lua_module](https://github.com/openresty/lua-nginx-module). I think the changes you're making are releated to [ngx_stream_lua_module](https://github.com/openresty/stream-lua-nginx-module).
> 
> @xynicole @dengsh12 Should we support this module by merging the flags into `moduleLuaDirectives` or do we want to model it as a separate optional module?

I thought in a separate file is easier for management. It seems they are registered as two modules in fact? If we merge into one file it comes to a new special case for the code generator. The relationships among `ParseOptions.DirectiveSources` is `or` so put them in two files won't hurt any valid config files.

PR #104, opened 2024-07-10
limit_req_zone with 4 arguments not being analyzed

oliveromahony:
### Proposed changes

Limit request zone has an issue with not being throwing: invalid number of arguments in \"limit_req_zone\" directive error despite being 4 arguments. 

### Checklist

### DIFF
--- a/analyze_test.go
+++ b/analyze_test.go
@@ -620,15 +620,6 @@ func TestAnalyze_limit_req_zone(t *testing.T) {
 			blockCtx{"http"},
 			false,
 		},
-		// "limit_req_zone ok stream": {
-		// 	&Directive{
-		// 		Directive: "limit_req_zone",
-		// 		Args:      []string{"$binary_remote_addr", "zone=one:10m", "rate=1r/s", "sync"},
-		// 		Line:      5,
-		// 	},
-		// 	blockCtx{"stream"},
-		// 	false,
-		// },
 	}
 
 	for name, tc := range testcases {
--- a/analyze_test.go
+++ b/analyze_test.go
@@ -603,7 +603,6 @@ func TestAnalyze_nap_app_protect_enable(t *testing.T) {
 	}
 }
 
-//nolint:funlen
 func TestAnalyze_limit_req_zone(t *testing.T) {
 	t.Parallel()
 	testcases := map[string]struct {
--- a/analyze_test.go
+++ b/analyze_test.go
@@ -604,6 +604,53 @@ func TestAnalyze_nap_app_protect_enable(t *testing.T) {
 }
 
 //nolint:funlen
+func TestAnalyze_limit_req_zone(t *testing.T) {
+	t.Parallel()
+	testcases := map[string]struct {
+		stmt    *Directive
+		ctx     blockCtx
+		wantErr bool
+	}{
+		"limit_req_zone ok http": {
+			&Directive{
+				Directive: "limit_req_zone",
+				Args:      []string{"$binary_remote_addr", "zone=one:10m", "rate=1r/s", "sync"},
+				Line:      5,
+			},
+			blockCtx{"http"},
+			false,
+		},
+		// "limit_req_zone ok stream": {
+		// 	&Directive{
+		// 		Directive: "limit_req_zone",
+		// 		Args:      []string{"$binary_remote_addr", "zone=one:10m", "rate=1r/s", "sync"},
+		// 		Line:      5,
+		// 	},
+		// 	blockCtx{"stream"},
+		// 	false,
+		// },
+	}
+
+	for name, tc := range testcases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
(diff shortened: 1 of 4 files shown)


analyze_test.go
```diff
@@ -603,6 +603,43 @@ func TestAnalyze_nap_app_protect_enable(t *testing.T) {
...
+			false,
+		},
+	}
+
+	for name, tc := range testcases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			err := analyze("nginx.conf", tc.stmt, ";", tc.ctx, &ParseOptions{
+				DirectiveSources: []MatchFunc{NgxPlusLatestDirectivesMatchFn, AppProtectWAFv4DirectivesMatchFn},
```
ryepup, line 629: I think this test would pass without any other changes if we use a different source

```suggestion
				DirectiveSources: []MatchFunc{NgxPlusR31DirectivesMatchFn, AppProtectWAFv4DirectivesMatchFn},
```

`NgxPlusLatestDirectivesMatchFn` thinks `limit_req_zone` has a mask of `ngxHTTPMainConf | ngxConfTake3 | ngxConfTake2,`, but `NgxPlusR31DirectivesMatchFn` has the correct mask `ngxHTTPMainConf | ngxConfTake3 | ngxConfTake4`

ornj: Adding details from our conversation external to the PR. `Take4` is only correct for NGINX+ to allow the `sync` flag. OSS is `Take3`.

dengsh12:
The problem comes from a mismatch in a legacy version of code generator, the flag `NGX_CONF_TAKE4` of `keyval` and `limit_req_zone` mapped to `ngxConfTake2` wrongly. `analyze_nplus_latest_directives.go` may be override to the legacy version while I was dealing with branches. I've proposed another MR #105 to fix. Thanks for your findings!

For the `uint` part it seems we define the bitmasks using `untyped int` and they can be compatible with `uint`, maybe we don't need to change them.

oliveromahony:
> The problem comes from a mismatch in a legacy version of code generator, the flag `NGX_CONF_TAKE4` of `keyval` and `limit_req_zone` mapped to `ngxConfTake2` wrongly. `analyze_nplus_latest_directives.go` may be override to the legacy version while I was dealing with branches. I've proposed another MR #105 to fix. Thanks for your findings!
> 
> For the `uint` part it seems we define the bitmasks using `untyped int` and they can be compatible with `uint`, maybe we don't need to change them.

Happy to drop close this PR. Can we add tests to cover this scenario in  https://github.com/nginxinc/nginx-go-crossplane/pull/105 ?

dengsh12:
> > The problem comes from a mismatch in a legacy version of code generator, the flag `NGX_CONF_TAKE4` of `keyval` and `limit_req_zone` mapped to `ngxConfTake2` wrongly. `analyze_nplus_latest_directives.go` may be override to the legacy version while I was dealing with branches. I've proposed another MR #105 to fix. Thanks for your findings!
> > For the `uint` part it seems we define the bitmasks using `untyped int` and they can be compatible with `uint`, maybe we don't need to change them.
> 
> Happy to close this PR. Can we add tests to cover this scenario in #105 ?

Yeah! Root reason comes from a legacy version of the generator has a wrong mapping from `NGX_CONF_TAKE4` to `ngxConfTake2` and I added [one unit test](https://github.com/nginxinc/nginx-go-crossplane/blob/e00ed7072a655fa52237f7c1500e6f0aaec51318/internal/generator/generator_util_test.go#L112C19-L112C43) with all bitmask in nginx to the MR of generator  #101 

It would also be helpful to add an unit test for `limit_req_zone`. Your test case in this PR offers a very good example, could I add it to #105 ?

ryepup:
> It would also be helpful to add an unit test for `limit_req_zone`. Your test case in this PR is a very good example, could I add it to #105 ?

please do

oliveromahony:
> > > The problem comes from a mismatch in a legacy version of code generator, the flag `NGX_CONF_TAKE4` of `keyval` and `limit_req_zone` mapped to `ngxConfTake2` wrongly. `analyze_nplus_latest_directives.go` may be override to the legacy version while I was dealing with branches. I've proposed another MR #105 to fix. Thanks for your findings!
> > > For the `uint` part it seems we define the bitmasks using `untyped int` and they can be compatible with `uint`, maybe we don't need to change them.
> > 
> > 
> > Happy to close this PR. Can we add tests to cover this scenario in #105 ?
> 
> Yeah! Root reason comes from a legacy version of the generator has a wrong mapping from `NGX_CONF_TAKE4` to `ngxConfTake2` and I added [one unit test](https://github.com/nginxinc/nginx-go-crossplane/blob/e00ed7072a655fa52237f7c1500e6f0aaec51318/internal/generator/generator_util_test.go#L112C19-L112C43) with all bitmask in nginx to the MR of generator #101
> 
> It would also be helpful to add an unit test for `limit_req_zone`. Your test case in this PR is a very good example, could I add it to #105 ?

That would be great

PR #105, opened 2024-07-10
Fix Nplus directives definition

dengsh12:
### Proposed changes

In #100 we provided support files for `ParseOptions.DirectiveSources`. For some reason `analyze_nplus_latest_directives.go` was override to a legacy version with mismatch from `NGX_CONF_TAKE4` to `ngxConfTake2`. This PR is to fix those directives(`keyval` and `limit_req_zone`). Thanks to @oliveromahony found this problem.

### Checklist

### DIFF
--- a/analyze_nplus_latest_directives.go
+++ b/analyze_nplus_latest_directives.go
@@ -695,8 +695,8 @@ var ngxPlusLatestDirectives = map[string][]uint{
 		ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxConfTake12,
 	},
 	"keyval": {
-		ngxHTTPMainConf | ngxConfTake3 | ngxConfTake2,
-		ngxStreamMainConf | ngxConfTake3 | ngxConfTake2,
+		ngxHTTPMainConf | ngxConfTake3 | ngxConfTake4,
+		ngxStreamMainConf | ngxConfTake3 | ngxConfTake4,
 	},
 	"keyval_zone": {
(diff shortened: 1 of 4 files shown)


ornj (approved):
Approved, but let's think about how we might be able to test this kind of scenario to see if there are any more errors (and prevent them in the future).

ryepup (approved):

dengsh12:
> Approved, but let's think about how we might be able to test this kind of scenario to see if there are any more errors (and prevent them in the future).

Sure! I think such problems boil down to some small errors in generator leading to error in only one or two directives. Current UT unlikely to discover such problems. I'm thinking of adding a huge "standard map" which includes most of directives to test the latest OSS/Nplus matchFunc(kind of like what I was doing in POC).

ornj:
@dengsh12 tests that will cover the generator tool will be useful when that lands. Maybe that's enough if there's enough coverage of the bitmasks.

ryepup:
> > Approved, but let's think about how we might be able to test this kind of scenario to see if there are any more errors (and prevent them in the future).
> 
> Sure! I think such problems boil down to some small errors in generator leading to error in only one or two directives. Current UT unlikely to discover such problems. I'm thinking of adding a huge "standard map" which includes most of directives to test the latest OSS/Nplus matchFunc(kind of like what I was doing in POC).

yet another case where a corpus of anonymized NGINX configs would be real handy

dengsh12:
update: added ut from @oliveromahony's PR #104

oliveromahony (approved):

PR #106, opened 2024-07-11
Provide `-filter` and `-override` with command line generator

dengsh12:
### Proposed changes

In #101, we provided a command line generator for support files. The generator creates a directives map and `matchFunc` based on the directives definition in the source code.

However, some directives in source codes are outdated or unofficial (like [`post_action` ](https://stackoverflow.com/questions/52603090/configure-nginx-to-make-a-background-request)in OSS). You may only want to include some official directives in output and we provide a `-filter` option to achieve this.

Some directives have different definition in crossplane vs source code(like `if` in OSS, we added `ngxConfExpr` to it). You may want to override the output and we provide a `-override` option to achieve this.

### Checklist


internal/generator/generator_util.go
```diff
@@ -184,12 +184,39 @@ func getMasksFromPath(path string) (directive2Masks map[string][]mask, err error
...
 
-func genFromSrcCode(codePath string, mapVariableName string, matchFnName string, writer io.Writer) error {
+func genFromSrcCode(codePath string, mapVariableName string, matchFnName string, writer io.Writer,
+	filter map[string]struct{}, override map[string][]mask) error {
 	directive2Masks, err := getMasksFromPath(codePath)
 	if err != nil {
 		return err
 	}
 
+	if filter != nil {
```
ryepup, line 194: recommend using `len(x) > 0` instead of `x != nil` when looking at maps and slices. `len` will return 0 for nils or non-nil but empty values, and we usually want to treat them the same.

```suggestion
	if len(filter) > 0 {
```

internal/generator/generator_util.go
```diff
@@ -184,12 +184,39 @@ func getMasksFromPath(path string) (directive2Masks map[string][]mask, err error
...
+		dropDirectives := []string{}
+		for d := range directive2Masks {
+			if _, found := filter[d]; !found {
+				dropDirectives = append(dropDirectives, d)
+			}
+		}
+
+		for _, d := range dropDirectives {
+			delete(directive2Masks, d)
+		}
```
ryepup, line 204: we can mutate a map while iterating through it, no need for the extra slice

```suggestion
		for d := range directive2Masks {
			if _, found := filter[d]; !found {
				delete(directive2Masks, d)
			}
		}
```

cmd/generate/main.go
```diff
@@ -9,19 +9,44 @@ package main
...
-			"the path of source code your want to generate support from, it can be either a file or a directory. (required)")
+			"The path of source code your want to generate support from, it can be either a file or a directory. (required)")
+		filter   stringSliceFlag
+		override stringSliceFlag
 	)
+	flag.Var(&filter, "filter",
+		"A list of strings specifying the directives to include in the output. "+
+			"If not provided, all directives in the src-path will be included.\n"+
+			"An example is: -filter directive1 -filter directive2...(optional)")
+	flag.Var(&override, "override",
```
ryepup, line 41: the override and it's microformat is bugging me a little bit; `main.go`'s job is generally to convert the user input into the data that the rest of the codebase (`generator.Generate` wants, and with `override` we're kinda doing half of that here (CLI args to a `[]string`), and then the other half is inside `Generate` to convert each string to a `map[string][]mask` entry.

an alternate design:

- make a new `type Override struct` that implements [`TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler)
- change `generator.Generate` to accept a `[]Override` - it's expecting the caller to parse the microformat from CLI args into a `Override`
- make a `type overrideSliceFlag []Override` here similar to `stringSliceFlag`, to convert the CLI args to an `[]Override`, returning any parse errors from `overrideSliceFlag.Set` so it fails early

Thoughts?

cmd/generate/main.go
```diff
@@ -9,19 +9,44 @@ package main
...
+		"A list of strings, used to override the output. "+
+			"It should follow the format:{directive:mask1,mask2,...}\n"+
+			"An example is -override=log_format:ngxHTTPMainConf|ngxConf2More,ngxStreamMainConf|ngxConf2More\n"+
```
ornj, line 44: Why does the caller need to provide `\n`?

internal/generator/generator_util.go
```diff
@@ -184,12 +184,39 @@ func getMasksFromPath(path string) (directive2Masks map[string][]mask, err error
...
-func genFromSrcCode(codePath string, mapVariableName string, matchFnName string, writer io.Writer) error {
+func genFromSrcCode(codePath string, mapVariableName string, matchFnName string, writer io.Writer,
+	filter map[string]struct{}, override map[string][]mask) error {
 	directive2Masks, err := getMasksFromPath(codePath)
 	if err != nil {
 		return err
 	}
 
+	if filter != nil {
+		dropDirectives := []string{}
```
ornj, line 195:
```suggestion
		var dropDirectives []string
```

Avoids allocation in the case that there are no values in `directive2Masks`. Remember that you can `append()` a `nil` slice as it will allocate as needed and return the new slice.

internal/generator/generator_util.go
```diff
@@ -201,3 +228,52 @@ func genFromSrcCode(codePath string, mapVariableName string, matchFnName string,
...
+		return nil
+	}
+
+	rv := map[string]struct{}{}
+	for _, str := range strs {
+		rv[str] = struct{}{}
+	}
+
+	return rv
+}
```
ornj, line 243: You created a special `[]string` type to collect filters, could that have been a `map[string]struct{}` to begin with?

```go
type AllowedDirectives struct {
	directiveNames map[string]struct{}
}

func (a *AllowedDirectives) Set(s string) error {
	if a.directiveNames == nil {
		a.directiveNames := make(map[string]struct{})
	}
	a.directiveNames[s] = struct{}{}
	return nil
}
``

File: internal/generator/generator.go [hunk 7]
```diff
@@ -15,6 +15,10 @@ import (
...
 // sourcePath and its subdirectories, then output the corresponding directive
 // masks map named "directives" and matchFunc named "Match" via writer.
-func Generate(sourcePath string, writer io.Writer) error {
-	return genFromSrcCode(sourcePath, "directives", "Match", writer)
+func Generate(sourcePath string, writer io.Writer, filter []string, rawOverride []string) error {
+	override, err := getOverride(rawOverride)
+	if err != nil {
+		return err
+	}
+	return genFromSrcCode(sourcePath, "directives", "Match", writer, strList2Map(filter), override)
```
ornj, line 23: You created a special `[]string` type to collect filters, could that have been a `map[string]struct{}` to begin with?

```go
type AllowedDirectives struct {
	directiveNames map[string]struct{}
}

func (a *AllowedDirectives) Set(s string) error {
	if a.directiveNames == nil {
		a.directiveNames = make(map[string]struct{})
	}
	a.directiveNames[s] = struct{}{}
	return nil
}
```

internal/generator/generator_util.go
```diff
@@ -201,3 +228,52 @@ func genFromSrcCode(codePath string, mapVariableName string, matchFnName string,
...
+		// rawStr should follow the format: directive:bitmask00|bitmask01|...,bitmask10|bitmask11|...
+		subs := strings.Split(rawStr, ":")
+		if len(subs) != 2 {
+			return nil, fmt.Errorf("override parameter %s format wrong", rawStr)
+		}
+
+		directive := strings.TrimSpace(subs[0])
+		if directive == "" {
+			return nil, errors.New("directive name can't be empty")
+		}
```
ornj, line 263: Nit: Including the offending format/line might be helpful in the case the caller has provided multiple.

```suggestion
		if directive == "" {
			return nil, fmt.Errorf("directive name is empty: override=%s", rawStr)
		}
```

internal/generator/generator_util.go
dengsh12: My caution before was delete may change the order of the keys since there may be a re-hash. I just checked it is safe to delete in iteration for go map. Thanks for telling me and I will fix

internal/generator/generator_util.go
```diff
@@ -201,3 +228,52 @@ func genFromSrcCode(codePath string, mapVariableName string, matchFnName string,
...
+
+		if strings.TrimSpace(subs[1]) == "" {
+			return nil, errors.New("directive masks can't be empty")
+		}
+
+		for _, varNames := range strings.Split(subs[1], ",") {
+			directiveMask := mask{}
+			for _, varName := range strings.Split(varNames, "|") {
+				directiveMask = append(directiveMask, strings.TrimSpace(varName))
+			}
```
ornj, line 268: Nit: If you know the size of the slice needed, it's better to set it. It doesn't really matter that much since this is a tool that is only run occasionally but good practice anyway.

```go
names := strings.Split(varNames, "|")
if len(names) == 0 {
	continue
}

directiveMask := make(mask, len(names))
for i, name := range names {
	directiveMask[i] = strings.TrimSpace(name)
}
```

cmd/generate/main.go
dengsh12: Originally I thought this is just for `go run main.go --help`, and the new line provides a better reading experience. I will move `\n` to the begin of next line

cmd/generate/main.go
dengsh12: > main.go's job is generally to convert the user input into the data that the rest of the codebase (generator.Generate wants, and with override we're kinda doing half of that here (CLI args to a []string),

Yeah.. The reason we need this is: some directives have different definition in crossplane vs source code(like `if` in OSS, we added `ngxConfExpr` to it in crossplane.  `ngxConfExpr` is not in OSS). And we want to provide users with flexibilities to handle such cases. After another `go generate` tool available, this tool is likely to use at some third-party modules unknown by crossplane, and the override should be very small(at most one or two directives I think, no override for most cases). I thought that's okay?

> an alternate design:
make a new type Override struct that implements [TextUnmarshaler](https://pkg.go.dev/encoding#TextUnmarshaler)
change generator.Generate to accept a []Override - it's expecting the caller to parse the microformat from CLI args into a Override
make a type overrideSliceFlag []Override here similar to stringSliceFlag, to convert the CLI args to an []Override, returning any parse errors from overrideSliceFlag.Set so it fails early

Yeah I think you are right it should be done at the same layer as `main.go`. Applied in latest commit

cmd/generate/main.go
dengsh12: Fixed: split "\n" out

internal/generator/generator_util.go
dengsh12: fixed

internal/generator/generator.go
dengsh12: Fixed: add `filterFlagReader` in `cmd_util.go`

internal/generator/generator_util.go
dengsh12: fixed: return the raw override in error

internal/generator/generator_util.go
dengsh12: > Nit: If you know the size of the slice needed, it's better to set it.

Get it! Applied

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,82 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/nginxinc/nginx-go-crossplane/internal/generator"
+)
+
+type filterFlagReader struct {
```
ornj, line 10: Nit: the name "reader" makes me think of `io.Reader` which this does not implement. You could just called it `filters`?

cmd/generate/main.go
ornj: @ryepup Do you have an example of how you would make use of the `TextUnmarshaler` interface here? I know [`flag.TextVar`](https://pkg.go.dev/flag#TextVar) accepts this type. How are you proposing he make use of the `TextUnmarshaler` interface with the need to `Set()` into the slice type? 

Or, is the purpose in implementing `TextUnmarshaler` to follow an established pattern that perhaps makes this type more usable else where?

internal/generator/generator_util.go
dengsh12: Applied: fix the length of directiveMask, add one error and corresponding ut

cmd/generate/cmd_util.go
dengsh12: applied: filterFlagReader -> filterFlag

cmd/generate/main.go
ryepup: I was thinking something very similar to what was written in the latest update, still using `flag.Var` to accumulate a slice (`TextVar` didn't seem to handle slices), and `UnmarshalText` to parse each individual value:

```golang
func (overrides *overrideParserList) Set(value string) error {
	var parser overrideParser
	err := parser.UnmarshalText([]byte(value))
	if err != nil {
		return fmt.Errorf("invalid override %q: %w", value, err) // <-- I'd expand the error returned to include the bad value
	}

	*overrides = append(*overrides, parser)
	return nil
} 
```

Moving parse errors into the CLI parsing part of the program removes some possible errors later in the main generate code path, and that "fail fast" was my primary motivation. 

`TextUnmarshaler` seemed to fit since the CLI arg is a textual representation of an override, but could just as easily be a standalone constructor func like `ParseOverride(string) (Override, error)`.

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (overrides *overrideParserList) String() string {
+	return fmt.Sprintf("%v", *overrides)
+}
+
+func (overrides *overrideParserList) Set(value string) error {
+	var parser overrideParser
+	err := parser.UnmarshalText([]byte(value))
+	if err != nil {
+		return err
```
ryepup, line 80: nit: we can include the raw value here and simplify some of the error building in `UnmarshalText`

```suggestion
		return fmt.Errorf("invalid override %q: %w", value, err)
```

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	reader.filter[value] = struct{}{}
+	return nil
+}
+
+func (reader *filterFlag) String() string {
+	return fmt.Sprintf("%v", *reader)
+}
+
+type overrideParser struct {
```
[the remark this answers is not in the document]
dengsh12: Fixed: rename overrideParser -> overrideItem

cmd/generate/main.go
```diff
@@ -18,10 +18,27 @@ import (
...
+			"An example is -override=log_format:ngxHTTPMainConf|ngxConf2More,ngxStreamMainConf|ngxConf2More"+"\n"+
+			`To use | and , in command line, you may need to enclose your input in quotes, i.e. -override="directive:mask1,mask2,...". (optional)`)
+
 	flag.Parse()
-	err := generator.Generate(*sourceCodePath, os.Stdout)
+
+	override := map[string][]generator.Mask{}
+	for _, parser := range overrideParsers {
+		override[parser.directive] = parser.masks
+	}
```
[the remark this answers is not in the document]
dengsh12: Fixed: Use `Set` method directly update the map

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,85 @@
...
+	directive string
+	masks     []generator.Mask
+}
+
+//nolint:mnd
+func (override *overrideParser) UnmarshalText(text []byte) error {
+	rawOverride := string(text)
+
+	// rawStr should follow the format: directive:bitmask00|bitmask01|...,bitmask10|bitmask11|...
+	subs := strings.Split(rawOverride, ":")
```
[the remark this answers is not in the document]
dengsh12: Fixed: use strings.Cut for spliting into two parts

cmd/generate/cmd_util.go
dengsh12: Fixed: changed error return format

cmd/generate/cmd_util_test.go
```diff
@@ -0,0 +1,97 @@
...
+//nolint:funlen
+func TestOverrideParser(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name     string
+		input    string
+		expected overrideItem
+		wantErr  bool
+	}{
+		{
```
ryepup, line 26: nit: there's another idiom for table-driven tests that uses a `map` to specify the name of the test case, I think that reads a little better:

```golang
	testcases := map[string]struct{
		input    string
		expected overrideItem
		wantErr  bool
	}{
		"normalFormat_pass": {
			input: "location:ngxHTTPMainConf|ngxConfTake12,ngxStreamMainConf",
			expected: overrideItem{
				directive: "location",
				masks: []generator.Mask{
					{"ngxHTTPMainConf", "ngxConfTake12"},
					{"ngxStreamMainConf"},
				},
			},
		},
		// <other test cases omitted for brevity>
	}
	for name, tc := range testcases{
		tc := tc
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			
			// <test omitted for brevity>
		})
	}}
```

cmd/generate/cmd_util_test.go
```diff
@@ -0,0 +1,97 @@
...
+			wantErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			var parser overrideItem
+			err := parser.UnmarshalText([]byte(tc.input))
```
ryepup, line 78: nit: naming: Golang likes short variables names (see https://go.dev/wiki/CodeReviewComments#variable-names), in tests you'll often see two variables `want` and `got` to represent what the test wanted to the code to return, and what was actually returned.

Recommend using something besides "parser"

```suggestion
			var got overrideItem
			err := got.UnmarshalText([]byte(tc.input))
```
or maybe `o`
```suggestion
			var o overrideItem
			err := o.UnmarshalText([]byte(tc.input))
```

cmd/generate/cmd_util_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+			if tc.wantErr && err == nil {
+				t.Fatal("expected error, got nil")
+			}
+
+			// If the testcase wants an error and there is an error, skip the output file validation.
+			// Output makes no sense when there is an error.
+			if err != nil {
+				return
+			}
```
ryepup, line 92: nit: if we're using the `require` library, it has some helpers to assert around errors. That can let us eliminate the leftover `if err != nil`

```suggestion
			if tc.wantErr {
				require.Error(t, err)
			} else {
				require.NoError(t, err)			
			}
```

ryepup (approved):

cmd/generate/cmd_util_test.go
dengsh12: Oh yeah, this one is better and align with other ut in crossplane. Applied.

cmd/generate/cmd_util_test.go
dengsh12: Get it. Applied

cmd/generate/cmd_util_test.go
dengsh12: Awesome! Thanks for telling me this. Applied

cmd/generate/cmd_util_test.go
ornj: @ryepup Many crossplane tests use a slice, but I agree that a `map` is better. You start to see hints of this when it becomes necessary to add a "name" or other key to the testcase struct.

@dengsh12 A more important benefit to using a `map` vs slice is that Go randomizes the order of keys when ranging over the `map`. I believe-- I do not have documentation that states this as fact on hand at the moment-- this is to dissuade anyone from getting used to maps being ordered in the event the internal details are changed. You can generally use this to your benefit as it will randomize the order the tests are run which can help to detect any unintentional shared memory that leaks through tests in the form of test flake.

README.md
```diff
@@ -74,7 +74,9 @@ func main() {
 ```
 
 # Generate support for third-party modules
-This is an example that takes the path of a third-party module source code to generate support for it. Assume the source code path of that module is `./src`. You can call `go run cmd/generate/main.go ./src`. The stdout will be like
+This is a simple example that takes the path of a third-party module source code to generate support for it. For detail usage of the tool, please run
```
ornj, line 77:
```suggestion
This is a simple example that takes the path of a third-party module source code to generate support for it. For detailed usage of the tool, please run
```

README.md
```diff
@@ -74,7 +74,9 @@ func main() {
...
+`go run ./cmd/generate/ --help`
+Assume the source code path of that module is `./src`. You can call `go run ./cmd/generate/ --src-path=./src`. The stdout will be like
```
ornj, line 79:
```suggestion
`go run ./cmd/generate/ --help`. Assuming the source code path of that module is `./src`, you can call `go run ./cmd/generate/ --src-path=./src`. The output will be similar to:
```

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+	if reader.filter == nil {
+		reader.filter = make(map[string]struct{})
+	}
+	reader.filter[value] = struct{}{}
+	return nil
+}
+
+func (reader *filterFlag) String() string {
+	return fmt.Sprintf("%v", *reader)
+}
```
ornj, line 32: Nit: no longer a "Reader"

```suggestion
func (f *filterFlag) Set(value string) error {
	if f.filter == nil {
		f.filter = make(map[string]struct{})
	}
	f.filter[value] = struct{}{}
	return nil
}

func (f *filterFlag) String() string {
	return fmt.Sprintf("%v", *f)
}
```

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+
+type overrideItem struct {
+	directive string
+	masks     []generator.Mask
+}
+
+func (item *overrideItem) UnmarshalText(text []byte) error {
+	rawOverride := string(text)
```
ornj, line 40: Nit: The [`bytes` package](https://pkg.go.dev/bytes) may offer all of the functions from the `strings` package that you may be able to use to avoid the (very small) cost of casting `[]byte` to `string`

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+	// rawStr should follow the format: directive:bitmask00|bitmask01|...,bitmask10|bitmask11|...
+	directive, definition, found := strings.Cut(rawOverride, ":")
+	if !found {
+		return fmt.Errorf("colon not found, override=%s, ", rawOverride)
+	}
+	directive = strings.TrimSpace(directive)
+
+	item.directive = directive
+	if directive == "" {
+		return errors.New("directive name is empty")
```
ornj, line 51: Include the override to communicate which one is being rejected?

```go
return fmt.Errorf("directive name is empty: override=%s", rawOverride)
```

I think I would do this for most errors where it would be helpful to see the specific input being rejected.

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+	masks     []generator.Mask
+}
+
+func (item *overrideItem) UnmarshalText(text []byte) error {
+	rawOverride := string(text)
+
+	// rawStr should follow the format: directive:bitmask00|bitmask01|...,bitmask10|bitmask11|...
+	directive, definition, found := strings.Cut(rawOverride, ":")
+	if !found {
+		return fmt.Errorf("colon not found, override=%s, ", rawOverride)
```
ornj, line 45: I'm not sure why you've added an extra comma and space?

```suggestion
		return fmt.Errorf("colon not found, override=%s", rawOverride)
```

If you're concerned that printing the override could look ambiguous, you could use `%q` to wrap the string in quotes.

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+	if ov == nil {
+		return "nil"
+	}
+	return fmt.Sprintf("%v", *ov)
+}
+
+func (ov *override) Set(value string) error {
+	if *ov == nil {
+		*ov = override{}
+	}
```
ornj, line 90: Nit: These patterns always give me pause. I think calling a method on a `nil` value is generally a programming error where a panic is warranted, but there are reasons to do this. I'm not sure that it makes sense here though. I think you want the caller to initialize the value.

cmd/generate/cmd_util.go
```diff
@@ -0,0 +1,99 @@
...
+	var item overrideItem
+	err := item.UnmarshalText([]byte(value))
+	if err != nil {
+		return fmt.Errorf("invalid override %s, reason:%w", value, err)
```
ornj, line 94: Errors more commonly follow this idiom:

```suggestion
		return fmt.Errorf("invalid override %s: %w", value, err)
```

ornj (approved):

README.md
dengsh12: Applied

cmd/generate/cmd_util.go
dengsh12: Sure! Applied. Should've checked all the usages after rename...

cmd/generate/cmd_util.go
dengsh12: We moved it to `Set` to simplify the errors in item.UnmarshalText
https://github.com/nginxinc/nginx-go-crossplane/pull/106#discussion_r1678144644
```go
func (ov *override) Set(value string) error {
	if *ov == nil {
		*ov = override{}
	}
	var item overrideItem
	err := item.UnmarshalText([]byte(value))
	if err != nil {
		return fmt.Errorf("invalid override %s:%w", value, err)
	}

	(*ov)[item.directive] = item.masks
	return nil
}
```

cmd/generate/cmd_util.go
dengsh12: https://github.com/nginxinc/nginx-go-crossplane/pull/106#discussion_r1681359367. Sorry for omitting removal of rawOverride output here. Just removed

cmd/generate/cmd_util.go
dengsh12: I went in this way because most of the situations user doesn't provide a `-override`, lazy initialization can avoid unnecessary allocation(and `Set` is not run in parallel so lazy initialization is safe here).

cmd/generate/cmd_util.go
dengsh12: Get it, applied.

PR #109, opened 2024-07-17
`go generate` support for njs/headersmore

dengsh12:
### Update: apply json config, merged with #112 
We plan to provide a tool to update supporting files(`analyze_*_directives.go`) by `go generate`. This MR is to provide `go generate` update for njs/headersmore.

`go generate` only targets for dynamic modules(or OSS, NPlus) we want to provide long-lasting supports for. To provide temporary support for a module, please refer to #101 

Here are the detailed changes:
- Add a shell script to fetch code from github and generate corresponding support files. It would make adding new dynamic modules or updating current existing support files easy.
- Add `.gen.go` suffix to generated files and ignore them in `golangci-lint`
- Add `go:generate` comments for njs/headersmore in `analyze.go`, now we can update `analyze_headersMore_directives.gen.go` and `analyze_njs_directives.gen.go` by `go generate`

### Checklist


scripts/generate/public_git.sh
```diff
@@ -0,0 +1,92 @@
...
+fi
+
+echo "start generate from $url"
+tmp=$(mktemp -d)
+git clone "$url" "$tmp" --depth 1 --branch "$branch"
+
+generate_command="go run ./cmd/generate --src-path=$tmp --directive-map=$directive_map --match-func=$match_func"
+generate_command="$generate_command > $path"
+
+eval "$generate_command"
```
UnwashedMeme, line 88: Generally should avoid eval unless there is a good reason to.

```suggestion
go run ./cmd/generate "--src-path=$tmp" "--directive-map=$directive_map" "--match-func=$match_func" > $path
```
should do just fine, no?

If you want to make this a little neater and avoid long lines the pattern I like (requires /bin/bash at the top) is:
```suggestion
genArgs=(
  "--src-path=$tmp"
  # line comments are ok here to describe what you're doing
  "--directive-map=$directive_map"
  "--match-func=$match_func"
)

go run ./cmd/generate "${genArgs[@]}" > "$path"
```

genArgs is a bash array that we then splat (with quoting intact) onto the run command. The really nice thing about this array construct is that it doesn't require the `\` to continue lines that makes comments not work and is error prone as you add/remove lines.

scripts/generate/public_git.sh
```diff
@@ -0,0 +1,92 @@
...
+EOF
+}
+
+while [ ! $# -eq 0 ]; do
+    case "$1" in
+        --help | -h)
+            help
+            exit 0
+            ;;
+        --directive_map | -d)
```
UnwashedMeme, line 33: the help function says `--directive-map` instead of `--directive_map`, which is right? same for match_func

scripts/generate/public_git.sh
UnwashedMeme: Another benefit of the bash array is that you can append to it, e.g. `genArgs+=(--directive-map=foo)`

With that, and the fact that it looks like a lot of the args are being passed through to cmd/generate can you do more of the error handling in there?  

E.g. if the the case statement was
```bash
    --match_func | -m)
            genArgs+=("--match-func=$2")
            shift
            ;;
```
and if the cmd/generate doesn't get a match function _it_ raises the error?

cmd/generate/main.go
```diff
@@ -19,6 +19,8 @@ func main() {
 	var (
 		sourceCodePath = flag.String("src-path", "",
 			"The path of source code your want to generate support from, it can be either a file or a directory. (required)")
+		directiveMapName  = flag.String("directive-map", "directives", `Name of the generated map variable. "Directives" by default. (optional)`)
+		matchFuncName     = flag.String("match-func", "Match", `Name of the generated matchFunc. "Match" by default. (optional)`)
```
UnwashedMeme, line 23: If these _need_ to be provided (as the shell script appears to enforce) then make them required here and cmd/generate now works better when invoked outside the shell script

scripts/generate/public_git.sh
dengsh12: > genArgs is a bash array that we then splat (with quoting intact) onto the run command.

Cool! I didn't know that before. Thanks for telling me this. You are right, the reason I used `eval` was I don't want long line and bash array seems a perfect solution. I will applied it.

>With that, and the fact that it looks like a lot of the args are being passed through to cmd/generate can you do more of the error handling in there?
E.g. if the the case statement was
    --match_func | -m)
            genArgs+=("--match-func=$2")
            shift
            ;;
and if the cmd/generate doesn't get a match function it raises the error?

I thought I already did the check here? Just after the arguments reading part

```bash
if [ "$url" = "" ]; then
    echo "url can't be empty"
    exit 1
fi

if [ "$path" = "" ]; then
    echo "path can't be empty"
    exit 1
fi

if [ "$directive_map" = "" ]; then
    echo "directive-map can't be empty"
    exit 1
fi

if [ "$match_func" = "" ]; then
    echo "match-func can't be empty"
    exit 1
fi
```

analyze.go
```diff
@@ -7,6 +7,9 @@
 
 package crossplane
 
+//go:generate sh ./scripts/generate/public_git.sh --url https://github.com/openresty/headers-more-nginx-module.git -p ./analyze_headersMore_directives.gen.go -m HeadersMoreDirectivesMatchFn -d moduleHeadersMoreDirectives
```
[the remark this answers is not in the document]
dengsh12: Oh there are some considerations for OSS and Lua in next MR. In next MR we will add OSS and Lua into `go generate`, and the arguments(`-filter` and `-override`) for `./cmd/generate` will be long. Here the command line for lua:
```bash
go run ./cmd/generate --src-path=/var/folders/pz/_91zbp7d7fb4mpclbtmzw4rr0000gp/T/tmp.uc45dimtKO --directive-map=moduleLuaDirectives --match-func=LuaDirectivesMatchFn -filter=lua_fake_shm  -override="content_by_lua_block:ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="rewrite_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="ssl_certificate_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxConfTake1" -override="ssl_session_fetch_by_lua_block:ngxHTTPMainConf | ngxConfTake1" -override="ssl_session_store_by_lua_block:ngxHTTPMainConf | ngxConfTake1" -override="set_by_lua_block:ngxHTTPSrvConf | ngxHTTPSifConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake2" -override="ssl_client_hello_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxConfTake1" -override="body_filter_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="init_by_lua_block:ngxHTTPMainConf | ngxConfTake1" -override="init_worker_by_lua_block:ngxHTTPMainConf | ngxConfTake1" -override="header_filter_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="log_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="access_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConfTake1" -override="balancer_by_lua_block:ngxHTTPUpsConf | ngxConfTake1" -override="server_rewrite_by_lua_block:ngxHTTPMainConf | ngxHTTPSrvConf | ngxConfTake1" -override="exit_worker_by_lua_block:ngxHTTPMainConf | ngxConfTake1" > ./analyze_lua_directives.go
```
We don't want this as a long line of `go:generate` comment. We plan to put the `-filter` and `-override` arguments in a separate shell script, or in a file and let `public-git.sh` to read that file in next MR. So we put the part of invoking `./cmd/generate` here. And the purpose of `public_git.sh` is only called by `go generate` to update `.gen.go` files. I should add some comments to illustrate this.

cmd/generate/main.go
dengsh12: My origin thought was: `public_git.sh` only targets for updating the `.gen.go` files in crossplane so it is required(these two arguments should follow the pattern in other `.gen.go`). `./cmd/generate` may be called by some users just want to generate a temporary support for a third-party module, so I provide a default value just for their convenience. 
But I'm unsure if this difference(`required` in `public_git.sh`, `optional` in `./cmd/generate`) will cause maintainers feel confused. Appreciate if you can provide some insights here.

scripts/generate/public_git.sh
dengsh12: Oh correct. Sorry for such minor error

scripts/generate/public_git.sh
UnwashedMeme: yeah the array-splat is one of my favorite bash tricks, e.g. look at the use of "opts" separately in most functions in https://gitlab.com/f5/nginx/nginxazurelb/azure-infrastructure/-/snippets/2371626

internal/generator/generator.go
```diff
@@ -14,7 +14,8 @@ import (
 // Generate receives a string sourcePath and an io.Writer writer. It will
 // extract all the directives definitions from the .c and .cpp files in
 // sourcePath and its subdirectories, then output the corresponding directive
-// masks map named "directives" and matchFunc named "Match" via writer.
-func Generate(sourcePath string, writer io.Writer, filter map[string]struct{}, override map[string][]Mask) error {
-	return genFromSrcCode(sourcePath, "directives", "Match", writer, filter, override)
+// masks map and matchFunc via writer.
+func Generate(sourcePath string, writer io.Writer, filter map[string]struct{}, override map[string][]Mask,
+	directiveMapName string, matchFuncName string) error {
+	return genFromSrcCode(sourcePath, directiveMapName, matchFuncName, writer, filter, override)
```
UnwashedMeme, line 20: strong stylistic nit-- if you need to make a function header multiple lines the return value should always be it's own line, never shared with params. See how at the moment `directiveMapName` and `return` are at the same indent level? That makes it harder to quickly identify what is function body vs parameters (in this case `return` is a pretty big hint though).

If we rearrange slightly:
```suggestion
func Generate(
        sourcePath string, 
        writer io.Writer, 
        filter map[string]struct{}, 
        override map[string][]Mask, 
        directiveMapName string, 
        matchFuncName string,
) error {
	return genFromSrcCode(sourcePath, directiveMapName, matchFuncName, writer, filter, override)
```

Now the func header is _very_ clearly delineated.  Occasionally you can group multiple params onto one line-- often with many params there are subgroups that go together.  Conversely many params is a bit of a smell itself.

UnwashedMeme:
I haven't worked in this repo yet so I'm coming in a bit blind; trying to assemble some pieces here:

> ./cmd/generate may be called by some users just want to generate a temporary support for a third-party module, so I provide a default value just for their convenience.

> there are some considerations for OSS and Lua in next MR. In next MR we will add OSS and Lua into go generate, and the arguments(-filter and -override) for ./cmd/generate will be long. Here the command line for lua: ...

> plan to put the -filter and -override arguments in a separate shell script

Your suggesting a user might do the git clone manually and call `cmd/generate` once and then build directly (never adding go:gen statements). That on that third party library they wouldn't need to do a directive-map/match-func, it's just that the lua and headersmore are super weird and need it. We're trying to include headers/lua as standard components to this repo.

### Idea 1 - lots of shell 
- public-git.sh should _only_ be responsible for doing the clone into temp
- makea  generate/lua.sh and generate/headersmore.sh that call public-git and then cmd/generate calls those, encode all the details

### idea 2 - more go
- public-git should still be only responsible for doing the clone 
- encode more of these options into cmd/generate and make a --module or --type or --preset flag to get there.

(i'll try and come back here with more detail later, but have to go now)

internal/generator/generator.go
dengsh12: Oh I see. Multiple lines seems better. 

> Conversely many params is a bit of a smell itself.

Yeah I'm quite unsure if I should pack them into a struct here... Is there a convention about the maximum number of params in our group?

dengsh12:
> I haven't worked in this repo yet so I'm coming in a bit blind; trying to assemble some pieces here:
> 
> > ./cmd/generate may be called by some users just want to generate a temporary support for a third-party module, so I provide a default value just for their convenience.
> 
> > there are some considerations for OSS and Lua in next MR. In next MR we will add OSS and Lua into go generate, and the arguments(-filter and -override) for ./cmd/generate will be long. Here the command line for lua: ...
> 
> > plan to put the -filter and -override arguments in a separate shell script
> 
> Your suggesting a user might do the git clone manually and call `cmd/generate` once and then build directly (never adding go:gen statements). That on that third party library they wouldn't need to do a directive-map/match-func, it's just that the lua and headersmore are super weird and need it. We're trying to include headers/lua as standard components to this repo.
> 
> ### Idea 1 - lots of shell
> * public-git.sh should _only_ be responsible for doing the clone into temp
> * makea  generate/lua.sh and generate/headersmore.sh that call public-git and then cmd/generate calls those, encode all the details
> 
> ### idea 2 - more go
> * public-git should still be only responsible for doing the clone
> * encode more of these options into cmd/generate and make a --module or --type or --preset flag to get there.
> 
> (i'll try and come back here with more detail later, but have to go now)

Oh I should've provided more context here. `go generate` is provided to update the `.gen.go` files in crossplane, or add a new dynamic module to crossplane. In the future all `analyze_*_directives.go` in crossplane will turn into `.gen.go` files. To update crossplane we need a MR each time. Once a `//go:generate` statement added to crossplane, the module is considered officially supported by crossplane.

`cmd/generate` targets for users with a third-party module, say `MD`. `MD` is not officially supported by crossplane(it may be just a module invented by the user, doesn't have a github link). The user may want to use crossplane to parse `MD`, but he doesn't want to open a MR for it. In this case he can call `cmd/generate` directly.

cmd/generate/main.go
```diff
@@ -19,6 +19,8 @@ func main() {
 	var (
 		sourceCodePath = flag.String("src-path", "",
 			"The path of source code your want to generate support from, it can be either a file or a directory. (required)")
+		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable.(required)`)
+		matchFuncName     = flag.String("match-func", "", `Name of the generated matchFunc.(required)`)
```
ornj, line 23:
```suggestion
		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable. (required)`)
		matchFuncName     = flag.String("match-func", "", `Name of the generated matchFunc. (required)`)
```

scripts/generate/public_git.sh
ornj, line 1: Is your intent to add a separate `.sh` file per data source "type", or could you infer from the `--url` option or allow setting a flag like `--git` or `--hg`?

scripts/generate/public_git.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "directive-map can't be empty"
+    exit 1
+fi
+
+if [ "$match_func" = "" ]; then
+    echo "match-func can't be empty"
+    exit 1
+fi
+
+echo "start generate from $url"
```
ornj, line 87: Nit: "No news is good news". I think don't print anything unless asked to do so. You may even wish to print the results to `stdout` and let the caller redirect to a file so you don't need the require the `--path` option? The appearance is more or less similar in the `//go:generate` directives but requires a little less shell script.

scripts/generate/public_git.sh
```diff
@@ -0,0 +1,101 @@
...
+
+genArgs=(
+  "--src-path=$tmp"
+  "--directive-map=$directive_map"
+  "--match-func=$match_func"
+)
+
+go run ./cmd/generate "${genArgs[@]}" > "$path"
+
+rm -rf "$tmp"
```
ornj, line 99: `tmp` will not be removed if there are any errors before reaching this line. Use `trap` to do clean up. Something like:

```sh
tmp=$(mktemp -d)
trap "rm -rf $tmp" EXIT
```

scripts/generate/public_git.sh
dengsh12: Originally I want to add a separate .sh file per data source "type". Like `git_source.sh` `hg_source.sh`. But currently no need for hg now, N+ will migrate to gitswam

scripts/generate/public_git.sh
dengsh12: > Nit: "No news is good news".

I see! Will change.

> You may even wish to print the results to stdout and let the caller redirect to a file so you don't need the require the --path option?

My caution for it is: when the script runs into error, we will still redirect into files. Not sure if it is acceptable

dengsh12: Updated: remove unnecessary echo output

scripts/generate/public_git.sh
dengsh12: Updated: added cleanup function and `trap cleanup`

PR #110, opened 2024-07-18
Provide three more custom option for command-line generator

dengsh12:
### Proposed changes

Provide `-directive-map-name, `-match-func-name` and `-match-func-comment` options for command-line generator. Allows users to specify the names of generated map and matchFunc, comment for matchFunc

### Checklist


cmd/generate/main.go
```diff
@@ -19,6 +19,8 @@ func main() {
 	var (
 		sourceCodePath = flag.String("src-path", "",
 			"The path of source code your want to generate support from, it can be either a file or a directory. (required)")
+		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable.(required)`)
+		matchFuncName     = flag.String("match-func", "", `Name of the generated matchFunc.(required)`)
```
UnwashedMeme, line 23: In the other MR you had mentioned wanting to provide a helpful default here. I like that idea but thought having a wrapper script enforce a variable was provided to be really weird.

If this is can reasonably have a default, and it's just that lua/headers need to provide a value then have those provide the value -- that's exactly what optional args are for.

What was your thinking for switching to required?

UnwashedMeme (approved):

cmd/generate/main.go
dengsh12: Originally I made `directive-map` and `match-func` optional just for convenience in this usercase: an user has a small dynamic module and he wants to use crossplane to parse configs contain that module(but we don't want him open a MR to add `//go:generate` for that module). In this case he can run `./cmd/generate/` directly, and pass the generated `MatchFunc` to `Parseoptions.DirectiveSources` and I thought providing default values for these two params save the user's time.

After you mentioned the problem, I agreed it's quite weird for maintainers to understand why we defined it as required in shell script but optional in `./cmd/generate`. Moreover, the default values not that make sense. `directive-map` is the name of the generated map, and `match-func` is the name of the generated matchFunc. Consider the former usercase, the user is generating a support file for a dynamic module(say it `module_a`), and the `directive-map` and `match-func` should be something relevant to `module_a`(like `aDirectives`, `MatchADirectives`). The union default values I provided before don't really make sense, and I'm also concerned `optional` may cause users to overlook them.

cmd/generate/main.go
```diff
@@ -19,6 +19,8 @@ func main() {
 	var (
 		sourceCodePath = flag.String("src-path", "",
 			"The path of source code your want to generate support from, it can be either a file or a directory. (required)")
+		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable.(required)`)
```
ornj, line 22:
```suggestion
		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable. (required)`)
```

Would this be more "self documenting" if the flag was `-variable` or `-variable-name`? The fact that the type is a map seems inconsequential.

cmd/generate/main.go
ornj:
```suggestion
		directiveMapName  = flag.String("directive-map", "", `Name of the generated map variable.(required)`)
		matchFuncName     = flag.String("match-func", "", `Name of the generated matchFunc. (required)`)
```

internal/generator/generator.go
```diff
@@ -11,10 +11,17 @@ import (
 	"io"
 )
 
+type GenerateConfig struct {
+	Filter           map[string]struct{}
+	Override         map[string][]Mask
+	DirectiveMapName string
+	MatchFuncName    string
+}
```
ornj, line 19: Nit: it would be good practice to add doc comments to this type to communicate what each of these config settings do.

internal/generator/testdata/expected/normalDefinition
```diff
@@ -27,9 +27,9 @@ var directives = map[string][]uint{
     },
 }
 
-// Match is a matchFunc for parsing an NGINX config that contains the
+// MyMatchFn is a matchFunc for parsing an NGINX config that contains the
 // preceding directives.
```
ornj, line 31: Nit: I think this testdata file is illustrating that the doc comment could be made more informative. As worded, the reader would need to look at the source to figure out what this function is doing.

Perhaps the generated functions are well named so it's not a big deal. Maybe there could be an optional flag that allows for some configuration of the doc string to say which nginx release or module this relates to?

ornj (approved):

internal/generator/generator.go
dengsh12: I see your point. But I thought the variable names here are clear? But it is just from my angle. If you felt unclear about any of them I can add comments. I'm trying to save some comments like this you've sent me before lol https://preview.redd.it/vs35fh1ruef51.jpg?width=640&crop=smart&auto=webp&s=3e1dd46e210f0499b9ac87187eefbde912bebdeb

internal/generator/testdata/expected/normalDefinition
dengsh12: > Perhaps the generated functions are well named so it's not a big deal.

Yeah we would name it like `LuaDirectivesMatchFn` in `//go:generate ...`. This is just test data, not exact dynamic module so the name is kind of not that descriptive

> Maybe there could be an optional flag that allows for some configuration of the doc string to say which nginx release or module this relates to?

In `go generate` we will name the `MatchFunc` well. The tool will only be called directly when user wants to support some dynamic modules locally without MR. I thought if they want they can write comments directly? It's easier than pass comments to a command-line argument

cmd/generate/main.go
dengsh12: At the beginning I also used `directive-map-name` but I'm not sure if it was too long lol. Changed to `directive-map-name` and `match-func-name` in latest commit

internal/generator/generator.go
ornj: I think the names are decent, but here's an example that I think adds value.

```go
// DirectiveMapName is the name assigned to the variable containing the directives
// discovered in the NGINX module source. The variable name generally starts with a
// lowercase to avoid export. Users will use the generated function named by 
// MatchFuncName to validate the module directives in nginx configurations.
DirectiveMapName string
```

`MatchFuncName` would similarly be documented to communicate that it should be pascal case to be exported.

internal/generator/testdata/expected/normalDefinition
ornj: > I thought if they want they can write comments directly? It's easier than pass comments to a command-line argument
 
You mean edit-- and re-edit-- when the file is generated?

Adding a `--comment` option is one way to do it. Another way would be to allow the caller to provide a module name or some other string that gets dropped into the template. Forgive the pseudo syntax that follows.

```go
// {{.MatchFuncName}} is a matchFunc for parsing an NGINX config that contains
// directives from the {{.ModuleName}} module
```

You could add some logic to it or format it differently to add a bit more flexibility.

cmd/generate/main.go
ornj: I think you ignored the bit where "map" isn't really necessary information to communicate, but that's OK I guess.

I'm not sure if you want to do anything to enforce that the resulting variable is unexported or just suggest to the caller that it shouldn't be.

dengsh12: > I'm not sure if you want to do anything to enforce that the resulting variable is unexported or just suggest to the caller that it shouldn't be.

I think it's just suggestion? User may really want the variable exported sometimes(e.g. they want to iterate though the map in other package)

dengsh12: Updated: add comments for `directive-map-name` and `match-func-name`

internal/generator/testdata/expected/normalDefinition
dengsh12: Updated: add `match-func-comment` option

internal/generator/generator.go
```diff
@@ -11,10 +11,33 @@ import (
...
-// Generate receives a string sourcePath and an io.Writer writer. It will
-// extract all the directives definitions from the .c and .cpp files in
-// sourcePath and its subdirectories, then output the corresponding directive
-// masks map named "directives" and matchFunc named "Match" via writer.
-func Generate(sourcePath string, writer io.Writer, filter map[string]struct{}, override map[string][]Mask) error {
-	return genFromSrcCode(sourcePath, "directives", "Match", writer, filter, override)
+type GenerateConfig struct {
+	Filter map[string]struct{}
+
+	Override map[string][]Mask
```
ornj, line 17: Nit: I think these could use comments too to at least explain what `string` is.

internal/generator/generator.go
```diff
@@ -11,10 +11,33 @@ import (
...
+
+	// MatchFuncName is the name assigned to the matchFunc generated by the generator.
+	// It should generally start with a uppercase to export.
+	// Users will use the generated function named by MatchFuncName
+	// to validate the module directives in nginx configurations.
+	MatchFuncName string
+
+	// MatchFuncComment is the comment appears above the generated MatchFunc.
+	// It may contain some information like what modules are included
+	// in the generated MatchFunc.
```
ornj, line 33: Should it start with `MatchFuncName` to follow Go doc comment conventions, or will you auto generate that part of the comment? What happens if this is left empty?

internal/generator/generator.go
dengsh12: Oh sure! These are useful information to add. Added in latest commit

internal/generator/generator.go
dengsh12: > Should it start with MatchFuncName to follow Go doc comment conventions

Sounds good. I added comments suggesting that it should start with MatchFuncName(but not enforced) in latest commit.

> What happens if this is left empty?

If it is empty, no code comments for `MatchFunc`. I added comments for this in latest commit.

ornj (approved):

PR #112, opened 2024-07-25
go generate support for oss/public modules

dengsh12:
### Proposed changes

We plan to provide a tool to update supporting files(`analyze_*_directives.go`) by `go generate`. This MR is to provide `go generate` update for oss and public modules(oss/lua/otel//njs/headersmore).

`go generate` only targets for dynamic modules(or OSS, NPlus) we want to provide long-lasting supports for. To provide temporary support for a module, please refer to #101 

Here are the detailed changes:
- Add a shell script(`./scripts/public_source_gen.sh`) to fetch code from github and generate corresponding support files. It would make adding new dynamic modules or updating current existing support files easy.
- Add `.gen.go` suffix to generated files and ignore them in `golangci-lint`
- Add `go:generate` comments for oss and public modules in `analyze.go`
- `./cmd/generate` can read from json file now
- Renaming of `MatchFunc` and directives map in `.gen.go` files.

### Checklist


.golangci.yml
```diff
@@ -42,6 +42,9 @@ run:
   timeout: 10m
   modules-download-mode: vendor
   skip-dirs-use-default: true
+  skip-files:
+    - '.*\.gen\.go$' 
+
```
ornj, line 47: I'm not 100% on why we would skip these files in the linter. Is there a reason the generated code files can't be made to pass?

analyze_headersMore_directives.gen.go
```diff
@@ -14,23 +15,23 @@
...
 }
 
+
 func HeadersMoreDirectivesMatchFn(directive string) ([]uint, bool) {
-	masks, matched := moduleHeadersMoreDirectives[directive]
-	return masks, matched
-}
+    m, ok := moduleHeadersMoreDirectives[directive]
+    return m, ok
+}
```
ornj, line 37: Nit: Can the code generator output a newline at the end of the generated code? That way you don't have this annoying icon in the diff and anyone who tries to print the end of the file to their terminal will be a little happier.

cmd/generate/main.go
```diff
@@ -8,58 +8,125 @@
...
+
+		content, err := io.ReadAll(file)
+		if err != nil {
+			return err
+		}
+
+		err = json.Unmarshal(content, &config)
+		if err != nil {
+			return err
+		}
```
ornj, line 53: Since you're reading from a stream (`os.File`) you can shorten this up by using `json.Decoder`. 🏌️ 

```go
f, err := os.Open(configPath)
if err != nil {
	return err
}

defer f.Close()

if err := json.NewDecoder(f).Decode(&config); err != nil {
	return err
}
```

There are some caveats about `json.Decoder` since it's decoding a stream that might contain more JSON objects, but I think in this case if we get the end of the first one then it's g2g.

With this it starts to look to me that you're better off keeping this stuff in `main()` so you can just fill the config object from the flags vs passing them as params to `generate()`.

scripts/generate/public_source_gen.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
ornj, line 1:
```suggestion
#!/usr/bin/env bash
```

More flexible

scripts/generate/public_source_gen.sh
```diff
@@ -0,0 +1,109 @@
...
+branch="master"
+url=""
+output_path=""
+sub_path=""
+config_path=""
+
+help() {
+    cat << EOF
+Generate support file from a public repo(currently only git).
+Only call me through //go:generate comments in project root to update .gen.go files.
```
ornj, line 22: Nit: I don't think the help text needs to say this. `//go:generate` is just a good way to do it as the `analyzer.go` file is basically documenting a dependency on those files.

scripts/generate/public_source_gen.sh
```diff
@@ -0,0 +1,109 @@
...
+
+branch="master"
+url=""
+output_path=""
+sub_path=""
+config_path=""
+
+help() {
+    cat << EOF
+Generate support file from a public repo(currently only git).
```
ornj, line 21: What about an existing directory on disk?

scripts/generate/public_source_gen.sh
```diff
@@ -0,0 +1,109 @@
...
+Only call me through //go:generate comments in project root to update .gen.go files.
+
+usage: $(basename "$0") [-b|--branch] [-o|--output-path] [-c|--config-path] [-s|--sub-path] [--url] [-h|--help]
+    -h  | --help                Display this message
+    -b  | --branch              Branch to checkout, defaults to "$branch". (optional)
+    -o  | --output-path         The path of the output file. (required)
+    -c  | --config-path         The path of json config file. Normally the json config file should be under nginx-go-crossplane/scripts/generate/configs.
+     The json config will be unmarshed into generator.GenerateConfig(at nginx-go-crossplane/internal/generator.go).
+     For the json config format, please refer to examples under nginx-go-crossplane/scripts/generate/configs,
+     or see generator.GenerateConfig directly. (required)
```
ornj, line 31: Why is this required?

```suggestion
    -c  | --config-path         The path of json config file. (required)
```

scripts/generate/public_source_gen.sh
```diff
@@ -0,0 +1,109 @@
...
+    -s  | --sub-path            The subfolder or file within the repository contains all the directives you want, whole repository by default. (optional)
```
ornj, line 32: Nit "folder" is a windows/UI concept. Use "directory".

I think this looks better:

```suggestion
    -p  | --path            Path to a directory of file in the repository containing
                            an nginx module. (optional)
```

ornj (changes requested):

scripts/generate/configs/oss_124_config.json
```diff
@@ -0,0 +1,44 @@
...
+        "header":{},
+        "devpoll_events":{},
+        "kqueue_events":{},
+        "postpone_gzipping":{},
+        "acceptex_read":{},
+        "iocp_threads":{},
+        "devpoll_changes":{},
+        "post_acceptex":{},
+        "open_file_cache_events":{},
+        "gzip_no_buffer":{}
```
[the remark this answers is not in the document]
ornj: Example: https://go.dev/play/p/qcrjzlIPGHc

.golangci.yml
dengsh12: I added it since the name of the generated directives map and matchFunc may be long and can't pass funlen lint.

analyze_test.go
```diff
@@ -294,7 +294,12 @@ func TestAnalyze_njs(t *testing.T) {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
-			err := analyze("nginx.conf", tc.stmt, ";", tc.ctx, &ParseOptions{})
+			err := analyze("nginx.conf", tc.stmt, ";", tc.ctx, &ParseOptions{
+				DirectiveSources: []MatchFunc{
+					NgxPlusLatestDirectivesMatchFn,
+					NjsDirectivesMatchFn,
+				},
```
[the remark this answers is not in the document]
dengsh12: I see. But if we change the names of current `MatchFuncs` for modules, current user may be influenced. Not sure if it is okay

cmd/generate/main.go
dengsh12: > Since you're reading from a stream (os.File) you can shorten this up by using json.Decoder. 

Cool! Sure!

> With this it starts to look to me that you're better off keeping this stuff in main() so you can just fill the config object from the flags vs passing them as params to generate().

I think so... The reason I extract a new `generate()` function is I want to write UT to satisfy the bot... If we put these in `main` we may need to set `os.Args` to test `main` and it can't be paralleled. Maybe we don't need UT  for this at all? The logic is quite simple, the functionality has been verified by `go generate`. And we have enough tests for `generator.generate`

.golangci.yml
ornj: I think you may be able to disable specific rules through the `golangci-lint` config if the file matches a pattern. You could match `.gen.go` files and disable linters that are hard to accommodate vs disable all linters.

analyze_test.go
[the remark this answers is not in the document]
ornj: Probably influenced any way, the API is not stable.

cmd/generate/main.go
ornj: I think the `main.go` is pretty minimal so up to you if you really think it needs tests. If nothing is happening other than taking command-line input and feeding it to a function that is otherwise well-tested, I don't know that we need it.

scripts/generate/public_source_gen.sh
dengsh12: Since we have `go run ./cmd/generate "${genArgs[@]}"` in shell script it can only run at project root. Do we want to put the path of ` ./cmd/generate` as an  param? It's more flexible but a new required para for the script

scripts/generate/public_source_gen.sh
dengsh12: I thought user should run `./cmd/generate` directly in this situation?

scripts/generate/public_source_gen.sh
dengsh12: Just want to let the script simple and have a unique way to configure for `go generate`. If we provide `--match-func-name`, `--directive-map-name`, `--filter`.... here like `./cmd/generate`, there maybe some modules doesn't have json config while some others have. A json config for every module seems more tidy?

.golangci.yml
dengsh12: Applied: only disable funlen for `.gen.go` files

analyze_headersMore_directives.gen.go
dengsh12: Applied: add a new line at the end of the generated code

analyze_test.go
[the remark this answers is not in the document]
dengsh12: Applied: rename generated `MatchFunc` and directives map name through json configs and regenerate

cmd/generate/main.go
```diff
@@ -8,58 +8,125 @@
...
+	directiveOverride map[string][]generator.Mask,
+	writer io.Writer) error {
+	var config generator.GenerateConfig
+	if configPath != "" {
+		file, err := os.Open(configPath)
+		if err != nil {
+			return err
+		}
+
+		defer func() {
```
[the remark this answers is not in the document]
dengsh12: Applied `defer f.Close()`

cmd/generate/main.go
dengsh12: Applied: delete unnecessary UT and `json.Decoder`

cmd/generate/main_test.go
```diff
@@ -0,0 +1,190 @@
...
+		`update the expected output of these tests, 
+ only use when the expected output is outdated and you are sure your output is correct`)
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+func getProjectRootAbsPath() (string, error) {
```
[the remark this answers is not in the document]
dengsh12: Unnecessary UT deleted

scripts/generate/configs/oss_124_config.json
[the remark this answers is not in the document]
dengsh12: Applied: create a new type `cmdConfig`, json config will be parsed to it and then parse to generate.GenerateConfig

scripts/generate/public_source_gen.sh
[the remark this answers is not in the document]
dengsh12: Originally I planed to add another `private_source_gen.sh` for NAP and NPlus. It seems we can add an `--enviroment-variable-name` to read the url of NAP and NPlus. I renamed it to `generate.sh`.

Applied: renamed to `generate.sh`

scripts/generate/public_source_gen.sh
dengsh12: Applied

scripts/generate/public_source_gen.sh
dengsh12: Applied. But a bit curious: is it okay we don't specify the default behavior here (scan the whole cloned repo)?

scripts/generate/public_source_gen.sh
ornj: That could be fine

scripts/generate/public_source_gen.sh
ornj: I think less tidy and annoying if nothing in the config file is actually needed to generate the module code.

scripts/generate/public_source_gen.sh
ornj: I might specify the behavior in the main body of `help()` and keep the description of the flags short.

analyze_headersMore_directives.gen.go
```diff
@@ -0,0 +1,37 @@
...
+    "more_set_headers": {
+        ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConf1More,
+    },
+    "more_set_input_headers": {
+        ngxHTTPMainConf | ngxHTTPSrvConf | ngxHTTPLocConf | ngxHTTPLifConf | ngxConf1More,
+    },
+}
+
+
+func MatchHeadersMore(directive string) ([]uint, bool) {
```
ornj, line 34: Nit: I think we should provide doc comments for all generated match functions included in the package. Perhaps it should mention which nginx module this is for (Even if it's obvious), maybe even including a module version?

cmd/generate/main.go
```diff
@@ -8,59 +8,126 @@
...
+		f, err := os.Open(*configPath)
+		if err != nil {
+			log.Println(err)
+			return
+		}
+
+		defer f.Close()
+
+		if err := json.NewDecoder(f).Decode(&rawConfig); err != nil {
+			log.Println(err)
```
ornj, line 86: I think you may want use `log.Fatal` which will call `os.Exit(1)` so that the command exits with a non-success error code. If the goal is to not prevent `f.Close()` from running then you could wrap the logic in a function that returns the error so that it can be logged with `log.Fatal` after the file is closed.

Either way, I think you want to make sure that the command does not exit with a `0` status code when there is a failure.

analyze_headersMore_directives.gen.go
dengsh12: Sure, applied

cmd/generate/main.go
dengsh12: > If the goal is to not prevent f.Close() from running then you could wrap the logic in a function that returns the error so that it can be logged with log.Fatal after the file is closed.

Yeah that's what I want. Cool idea, applied in latest commit.

scripts/generate/public_source_gen.sh
dengsh12: Applied: move the default behavior to `help()`

scripts/generate/public_source_gen.sh
dengsh12: I modified the params of `generate.sh`, kept them same as `./cmd/generate`. For those don't have `filter` and `override` I put them directly in `//go:generate` commands in `analyze.go` But now there are a lot of params(maybe it's okay? since we indeed have such many params).

scripts/generate/public_source_gen.sh
ornj: No, I don't think we need that.

cmd/generate/main.go
```diff
@@ -8,59 +8,135 @@
...
 
-	config := generator.GenerateConfig{
-		Filter:           filterflags.filter,
-		Override:         directiveOverride,
-		DirectiveMapName: *directiveMapName,
-		MatchFuncName:    *matchFuncName,
-		MatchFuncComment: *matchFnComment,
+	if config.MatchFuncName == "" {
+		log.Println("name of matchFunc can't be empty")
+		return
```
ornj, line 134: `log.Fatal`?

cmd/generate/main.go
```diff
@@ -8,59 +8,135 @@
...
 
-	if *directiveMapName == "" {
-		log.Fatal("directive-map can't be empty")
+	if *matchFnComment != "" {
+		config.MatchFuncComment = *matchFnComment
+	}
+
+	if *sourceCodePath == "" {
+		log.Fatal("src-path can't be empty")
+		return
```
ornj, line 124: fwiw, since `log.Fatal` calls `os.Exit(1)` you don't need the `return`.

scripts/generate/generate.sh
```diff
@@ -0,0 +1,139 @@
...
+if [ "$sub_path" = "" ]; then
+    git clone "$url" "$tmp" --depth 1 --branch "$branch" -q 1>&2
+else
+    (
+        git clone "$url" "$tmp" --no-checkout --depth 1 --branch "$branch" -q 1>&2 &&
+        # sparse-checkout doesn't support -q option
+        git -C "$tmp" sparse-checkout init --cone 1>&2  &&
+        git -C "$tmp" sparse-checkout set "$sub_path" 1>&2 &&
+        git -C "$tmp" checkout -q 1>&2
+    )
```
ornj, line 135: Nit: You can do `cd "$tmp"` between the `()` as the directory will only be changed between the parentheses as those create a subshell.

scripts/generate/public_source_gen.sh
ornj: I'm not sure why we're talking about the number of params in the generate directive. My comment was only that the config file should be optional. If you want to use it to keep line lengths short, you should use it.

scripts/generate/generate.sh
dengsh12: Awesome! Applied: ` cd "$tmp" ` after git clone

cmd/generate/main.go
dengsh12: Fixed: change omitted `log.Println` to `log.Fatal`

scripts/generate/public_source_gen.sh
dengsh12: Get it! Applied: use `--config-path` for all `go:generate` comments in `analyzed.go` to shorten line length, while `--config-path` is optional. Added documents to `./scripts/generate/generate.sh`

scripts/generate/public_source_gen.sh
dengsh12: Didn't change: retain the comments

analyze_nplus_latest_directives.go
```diff
@@ -2151,7 +2151,7 @@ var ngxPlusLatestDirectives = map[string][]uint{
 	},
 }
 
-func NgxPlusLatestDirectivesMatchFn(directive string) ([]uint, bool) {
+func MatchLatestNgxPlus(directive string) ([]uint, bool) {
```
ornj, line 2154: Can you add doc comments for the remaining match functions? If that's a different PR, that's OK too.

ornj: Also this should be `MatchNginxPlusLatest`

- Don't abbreviate "Nginx". The name is short enough as it is. We're not gaining anything and losing clarity.
- Follow the naming pattern esablished (`MatchModuleVersion`) so that the identifiers sort well. If you transpose the module name and version then the list looks unsorted and harder to scan in code editors as well as [pkg.go.dev](https://pkg.go.dev/github.com/nginxinc/nginx-go-crossplane)

dengsh12: Yeah since code comments will be included in `go generate`, I plan to add the corresponding comments when we have `go generate` comment for them.

cmd/generate/main.go
```diff
@@ -8,58 +8,130 @@
...
 import (
+	"encoding/json"
 	"flag"
 	"log"
 	"os"
 
 	"github.com/nginxinc/nginx-go-crossplane/internal/generator"
 )
 
+type cmdConfig struct {
```
ornj, line 19: I'm not sure why you need what looks like a duplicate of the `generator.GenerateConfig`. I think you can just have one type.

analyze_nplus_latest_directives.go
dengsh12: Applied: `Ngx` to `Nginx` in directive map variable name and MatchFunc, rename modules matchfunc to follow `MatchModuleVersion` pattern

cmd/generate/main.go
dengsh12: At our former discussion, https://github.com/nginxinc/nginx-go-crossplane/pull/112#discussion_r1693504383
We want to parse `filter` as an array first, then transfer it into a map.

ornj: I don't think you need another type to do this. The linked thread includes the [example I provided](https://go.dev/play/p/qcrjzlIPGHc) to accomplish this. You can decode the JSON with the array into the map type without an intermediate, duplicate struct.

dengsh12: Oh I see. Sorry I misunderstood the example before. Applied.

scripts/generate/configs/headersmore_config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "directiveMapName":"headersMoreDirectives",
+    "matchFuncName":"MatchHeadersMoreLatest",
+    "matchFuncComment":"MatchHeadersMoreLatest is a MatchFunc for the latest version of headersmore."
+}
```
ornj, line 5: Nit: new line at the end of these config files please?

ornj (approved):

scripts/generate/configs/headersmore_config.json
dengsh12: Oh sure! Applied

analyze.go
```diff
@@ -7,6 +7,25 @@
...
 package crossplane
 
+// Update for headersmore
+//go:generate sh -c "sh ./scripts/generate/generate.sh --url https://github.com/openresty/headers-more-nginx-module.git --config-path ./scripts/generate/configs/headersmore_config.json > ./analyze_headersMore_directives.gen.go"
+
+// Update for njs
+//go:generate sh -c "sh ./scripts/generate/generate.sh --url https://github.com/nginx/njs.git --config-path ./scripts/generate/configs/njs_config.json > ./analyze_njs_directives.gen.go"
+
+// Update for OSS, filter in config is the directives not in https://nginx.org/en/docs/dirindex.html but in source code.
+// Override in config is for the "if" directive. We create a bitmask ngxConfExpr for it in crossplane, which is not in source code.
```
valyria257, line 17: nit: do these comments about filter and override for each module make more sense in the corresponding json files? Or does that mess with the generator?

cmd/generate/main.go
```diff
@@ -8,58 +8,107 @@
...
+	}
+
+	if *sourceCodePath == "" {
+		log.Fatal("src-path can't be empty")
 	}
 
-	if *matchFuncName == "" {
-		log.Fatal("match-func can't be empty")
+	if config.DirectiveMapName == "" {
+		log.Fatal("name of directive map can't be empty")
```
valyria257, line 104: nit: use exact parameter name
```suggestion
		log.Fatal("directiveMapName can't be empty")
```

cmd/generate/main.go
```diff
@@ -8,58 +8,107 @@
...
 	}
 
-	config := generator.GenerateConfig{
-		Filter:           filterflags.filter,
-		Override:         directiveOverride,
-		DirectiveMapName: *directiveMapName,
-		MatchFuncName:    *matchFuncName,
-		MatchFuncComment: *matchFnComment,
+	if config.MatchFuncName == "" {
+		log.Fatal("name of matchFunc can't be empty")
```
valyria257, line 108: nit: use exact parameter name
```suggestion
		log.Fatal("matchFuncName can't be empty")
```

valyria257 (approved):

analyze.go
dengsh12: I also wanted in this way but sadly... json doesn't provide comments. I could only comment by adding keys like "filter_comment":"....." But seems a bit weird. 
I thought not deal breaker? Normally user will go to `analyze.go` to add & review `//go:generate` comments first.

valyria257: Yeah not a deal breaker :) comments in analyze.go works!

PR #116, opened 2024-07-31
go generate support for private modules(currently NAP)

dengsh12:
### Proposed changes

- Add NAP into `go generate`. To hide the private link from public repo we use environment variables. If in the future we want to add other private modules, we can go similar way.
- Add `.gen.go` suffix to generated NAP `analyze_appprotectWAF{version}_directives`
### Checklist


scripts/generate/generate.sh
```diff
@@ -56,7 +58,12 @@ usage: $(basename "$0") [-b|--branch] [-c|--config-path] [-d|--directive-map-nam
     
     -p  | --path                Path to a directory in the repository containing the source code of the nginx module. (optional)
     
-    --url                       Url used for git clone. (required)
+    --url                       Url used for git clone. You should provide url here or through --env-var-for-url.
+
+    -eu  | --env-var-for-url    Name of environment variable contains url. You should provide url here or through --url.
```
ornj, line 63: Instead of adding a second set of parameters, could you call the script like this?

```sh
./scripts/generate.sh --url $NAP_URL --branch $NAP_BRANCH
```

analyze.go
```diff
@@ -26,6 +26,16 @@ package crossplane
 // Otel provides its own config handler for some directives and they don't have context. Currently we don't support them.
 //go:generate sh -c "sh ./scripts/generate/generate.sh --url https://github.com/nginxinc/nginx-otel.git --config-path ./scripts/generate/configs/otel_config.json --branch main > ./analyze_otel_directives.gen.go"
 
+// Update for NAP v4 and v5.
+// NAP is a private module. Please ensure you have correct access and put the url
+// and branch of it in environment variable NAP_URL, NAP_V4_BRANCH, and NAP_V5_BRANCH.
+// Most of users don't have access, so we only run generate.sh when corresponding environment variables are set.
+// Override is for flag dirctives. NAP used ngxConfTake1 for flag directives, we change them to ngxConfFlag in crossplane.
+// NAP v4
+//go:generate sh -c "if [ -z $NAP_V4_BRANCH ] || [ -z $NAP_URL ] ; then exit 0; else sh ./scripts/generate/generate.sh --url $NAP_URL --config-path ./scripts/generate/configs/nap_v4_config.json --branch $NAP_V4_BRANCH --path ./src > analyze_appProtectWAFv4_directives.gen.go; fi"
```
ornj, line 35: Is this better than allowing the call to occur with an empty value? If any of the variables are unset then they will be empty. I'm wondering if the error returned from the script (something like saying the `--url` or `--branch` is required will give a better hint as to why the command is not working vs `exit 0` which makes it look like the command worked.

scripts/generate/configs/nap_v5_config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "directiveMapName":"appProtectWAFv5Directives",
+    "matchFuncName":"MatchAppProtectWAFv5",
+    "matchFuncComment":"MatchAppProtectWAFv5 is a MatchFunc for App Protect v5 module.",
+    "override":{
+        "app_protect_enable":[["ngxHTTPMainConf", "ngxHTTPSrvConf" , "ngxHTTPLocConf", "ngxConfFlag"]],
+        "app_protect_security_log_enable":[["ngxHTTPMainConf", "ngxHTTPSrvConf", "ngxHTTPLocConf", "ngxConfFlag"]]
+    }
+}
```
ornj, line 9: Nit: new lines please :-)

analyze.go
dengsh12: I assume most of users doesn't have access to NAP? If we allow the call with empty string, I'm afraid there's likely an error message `url can't be empty` and after user locates the error they would find they don't have access to NAP and just ignore it at last. 

What about going this way? I applied in latest commit. Added an `echo` to illustrate and `exit 1`
`//go:generate sh -c "if [ -z $NAP_V4_BRANCH ] || [ -z $NAP_URL ] ; then echo NAP v4 information missing, please set the environment variable NAP_URL, NAP_V4_BRANCH and ensure you have correct access && exit 1; else sh ./scripts/generate/generate.sh --url $NAP_URL --config-path ./scripts/generate/configs/nap_v4_config.json --branch $NAP_V4_BRANCH --path ./src > analyze_appProtectWAFv4_directives.gen.go; fi"
`

scripts/generate/configs/nap_v5_config.json
dengsh12: Oh! I must remember next time. Applied

ornj (approved):

analyze.go
dengsh12: Applied: suggest user without NAP access using `-skip` option

valyria257 (approved):

PR #117, opened 2024-08-01
NLB-5021: Update crossplane to handle GeoIP2 directives

xynicole:
### Proposed changes

https://github.com/leev/ngx_http_geoip2_module
Update crossplane to handle GeoIP2 directives

### Checklist


analyze_map.go
```diff
@@ -39,6 +39,7 @@ var mapBodies = map[string]mapParameterMasks{
 		defaultMasks: ngxConfTake1,
 	},
 	"geoip2": {
+		specialParameterMasks: map[string]uint{"auto_reload": ngxConfTake1},
 		defaultMasks: ngxConf1More,
```
ornj, line 43: Is this something that perhaps could be added to the generator in some form?

xynicole: It might be a little difficult because the source code doesn't include the bitmask
https://github.com/search?q=repo%3Aleev%2Fngx_http_geoip2_module%20auto_reload&type=code
`if (value[0].len == 11
            && ngx_strncmp(value[0].data, "auto_reload", 11) == 0) {
        if ((int) cf->args->nelts != 2) {
            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
                               "invalid number of arguments for auto_reload");
            return NGX_CONF_ERROR;
        }`

xynicole: Definitely! It would be great if the generator could have this capability. That way, we won't miss anything that's not explicitly stated in the README.

dengsh12: I took a look into source codes. It seems if we add a new field "value used for map-like directives" into generator, all of them still need to be manually configured. I failed to figure out a way to get `defaultMasks` or `specialParameterMasks` by scanning the source code. As long as the handler in source code aligns with ` char               *(*set)(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);`, the module author can write codes in any ways.

Given this I wonder:

1. Do we want to provide an interface for users to configure special bitmask for map-like directives? An user may have a third-party module not in crossplane, say it `md`. `md` also has a map-like directive `dir`. Currently the user can't parse it.
2. If so, do we want users to specify it through generator, or directly implement the interface?

dengsh12: Out of scope of current MR and opened an issue to discuss:https://github.com/nginxinc/nginx-go-crossplane/issues/119.

dengsh12 (approved):

ornj (approved):

PR #134, opened 2025-02-18
NLB-8186: identify directives as parameters in map blocks

valyria257:
### Proposed changes

 Crossplane parses map-like directives by adding the parameters in their bodies to the payload
as Directives. However, the parameters within the map blocks are not valid nginx directives;
and therefore, it is useful to discern which Directive types that crossplane generates are
nginx directives or map parameters.

### Checklist


types.go
```diff
@@ -38,13 +38,14 @@ type ConfigError struct {
...
-	Block     Directives `json:"block,omitempty"`
-	Comment   *string    `json:"comment,omitempty"`
+	Directive         string     `json:"directive"`
+	Line              int        `json:"line"`
+	Args              []string   `json:"args"`
+	File              string     `json:"file,omitempty"`
+	Includes          []int      `json:"includes,omitempty"`
+	Block             Directives `json:"block,omitempty"`
+	Comment           *string    `json:"comment,omitempty"`
+	MapBlockParameter bool       `json:"mapBlockParameter,omitempty"`
```
ryepup, line 48: do we need the getter? We could name this field `IsMapBlockParameter`?

valyria257: Updated to `IsMapBlockParameter` and removed the getter

ryepup (approved):
great commit messages, BTW!

xynicole (approved):

